PR #96, opened 2026-05-05
feat(integration): wire Ollama LLM factory with mock fallback

### LINKED ISSUE #59: Wire real Ollama LLM behind LLMService factory (mock fallback)
M1.5 / PR 6 — `integration/01-real-llm`. Wire Person A's real Ollama-backed LLM service behind the existing `LLMService` contract and keep the mock as offline fallback.

Includes:
- Add `OllamaLLMService implements LLMService` (Person A delivers the implementation; this PR is the UI-side wire-up + factory).
- `LLMServiceFactory` in `main.ts` selects between `OllamaLLMService` and `MockLLMService` based on a dev-only setting; default is `ollama`.
- Fallback path: if Ollama is unreachable on startup, fall back to mock and surface an Obsidian `Notice`.
- `MockLLMService` stays in the tree and CI as the offline/demo fallback.

### Acceptance
- With Ollama running, the chat streams real model output through the existing UI.
- With Ollama stopped, plugin starts cleanly, shows a Notice, and chat works against the mock.
- Switching the dev-only setting flips the factory without restart (or with a documented restart, whichever is simpler).

### Reference
`planning/ui-plugin-plan.md` §M1.5 (PR 6) and §Dependencies on other tracks.

iamaxeloberg:
## What
The plugin now checks if Ollama is running when it starts. If Ollama is available it uses it as normal. If not, it falls back to the mock LLM and shows a notice to the user. There is also a dev setting to force mock mode without needing Ollama at all.

## Why
Closes #59. Before this, starting the plugin without Ollama running would silently fail. Now offline and demo use works out of the box.

## How to test
- Start Obsidian with Ollama running — chat should work as normal with real model output
- Start Obsidian with Ollama stopped — a notice should appear and chat should still work using the mock
- \`npm test\` — all tests pass

Closes #59

### DIFF
--- a/__mocks__/obsidian.ts
+++ b/__mocks__/obsidian.ts
@@ -1,0 +1,14 @@
+export class Notice {
+  constructor(public message: string) {}
+}
+
+export class FileSystemAdapter {
+  getFullPath(path: string): string {
+    return path;
+  }
+}
+
+export class Plugin {}
+export class WorkspaceLeaf {}
+export class ItemView {}
+export class MarkdownRenderer {}
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,12 +1,15 @@
 import { Plugin, WorkspaceLeaf } from "obsidian";
 import { GemmeraChatView, VIEW_TYPE } from "./view";
 import { createServices, Services } from "./services";
+import { DEFAULT_SETTINGS, GemmeraSettings } from "./settings";
 
 export default class GemmeraPlugin extends Plugin {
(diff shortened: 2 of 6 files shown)


heimdallSWE (approved):
LGTM. Tests (115) pass, typecheck clean. Pushed a small docs commit clarifying that `llmBackend` is dev-only and not exposed in the UI. Mid-session Ollama re-check tracked in #98.

Issue #16, opened 2026-04-26
Eval: golden set + retrieval metrics + weekly regression check


heimdallSWE:
Build the eval harness from planning/rag.md "Evaluation: golden set and retrieval metrics".

- `evals/golden.jsonl` seeded with 30 `(question, ideal_note_paths, ideal_answer_summary)` examples.
- Metrics per run: Recall@10, Recall@30 (pre-rerank), MRR, citation correctness, latency P50/P95 for retrieval/rerank/generation/end-to-end.
- v1 faithfulness check is the rule-based version (every cited chunk exists in the retrieval payload). LLM-judge faithfulness is deferred to v2.
- Scheduled local job re-runs the set weekly, writes to a `metrics` table, app shows a small chart.
- Any regression of 10 percent or more in recall after a config change is flagged in the UI.

### Acceptance
- A single command runs the full eval set and prints metrics.
- A deliberate chunker regression in a test fixture trips the regression flag.
- Results are persisted so trend lines render in the UI.

iamaxeloberg:
Worth disambiguating: the eval that shipped (#85, #86, #87) is the **classifier eval** — intent-label precision/recall/F1 over `eval/classifier-golden/`. This issue is about **retrieval eval** — Recall@10, Recall@30, MRR, citation correctness, latency percentiles over question→chunk pairs. Different golden set, different metrics.

**Status:**
- Classifier eval infrastructure (runner, regression gate, CI) exists and can be a template for retrieval eval
- Retrieval golden set: not yet created
- Retrieval metrics computation: not implemented
- Weekly scheduled job + metrics table + UI chart: not implemented
- Faithfulness check (every cited chunk exists in retrieval payload): not implemented

Recommend keeping open, but probably v2 — depends on #14's rerank step landing first.

iamaxeloberg:
**Correction to my earlier comment.** Dev has `evals/golden/` with link-graph fixtures (`evals/golden/link-structure.jsonl` + 11 fixture markdown files) — that's distinct from the classifier eval. Whether the retrieval metrics (Recall@K, MRR) are computed by a runner is unverified; I only confirmed the fixtures exist. Re-evaluate scope after merging dev into `feature/classifier-milestone`.

Issue #14, opened 2026-04-26
Tools: query tool loop (search, rerank, answer, optional synthesis)


heimdallSWE:
Implement the on-question sequence from planning/rag.md "On question":

1. `search_notes` (hybrid)
2. `rerank_results`
3. `get_note_context` for the top-N if needed
4. Gemma answers grounded in the payload
5. Optional `create_synthesis_note` if the user wants the answer written back

### Acceptance
- Vault questions never bypass retrieval (enforced by a test that asserts at least one `retrieval` event per vault-tagged turn).
- Answers always include source paths in the UI payload.
- Synthesis notes carry the same `cowork` block + frontmatter contract as ingest output.

iamaxeloberg:
Partial implementation in `runAskPath` (`src/view.ts`):

**Done:**
- `search_notes` (vector search via `BinaryVectorStore` / `EmbeddingService`)
- Gemma answers grounded in retrieved payload (streaming)
- Source paths surfaced as citation chips (`appendCitations`)

**Missing:**
- `rerank_results` — bge-reranker-v2-m3 was planned but not wired; no rerank step
- `get_note_context` for top-N — current retrieval returns chunks directly, no separate context-expansion step
- `create_synthesis_note` — depends on the write path (#55 for direct-save, #62 for full dispatcher; #62 deferred to v2)
- Test asserting at least one retrieval event per vault-tagged turn

Recommend keeping open, but the rerank + synthesis pieces probably belong in v2. The minimum (search + answer + cite) ships today.

iamaxeloberg:
**Correction to my earlier comment.** I evaluated this only against `feature/classifier-milestone` and missed dev's implementation. PR #111 (merged to dev) ships the retrieval stack: `src/services/hybrid-retriever.ts`, `bm25-index-service.ts`, `payload-assembler.ts`, `query-orchestrator.ts`. Rerank schema exists at `schemas/gemma-tools/rerank_results.json`; need to verify whether the rerank step is wired in the orchestrator. Synthesis-note generation looks unimplemented in dev. Likely partial in dev — re-evaluate after merge.

SimonProjekt:
[...]

Issue #13, opened 2026-04-26
Tools: ingest tool loop (with near-duplicate branch)


heimdallSWE:
Implement the on-upload sequence from planning/rag.md "On upload":

1. `ingest_content`
2. backend extracts candidate note structure
3. `retrieve_similar_notes`
4. Gemma decides: new note, update existing, or link-only
5. `save_note`
6. `update_index`

The "near-duplicate found" branch must short-circuit to update-or-link without creating a duplicate file.

### Acceptance
- Re-uploading the same content does not create a second note; it surfaces the existing one.
- Updates preserve `cowork_managed = TRUE` invariants from planning/rag.md.
- Every step writes an `events` row.

Issue #10, opened 2026-04-26
Retrieval: structured payload assembler for Gemma context


heimdallSWE:
Build the compact payload described in planning/rag.md §"What to send to Gemma":

- note title, chunk text, section heading, note path
- why-matched tag, optional score
- optional neighboring note titles (1-hop)

The payload is the only retrieval-derived input the chat loop hands to Gemma.

### Acceptance
- Payload size stays bounded (configurable max chunks, default 8).
- Source paths are preserved end-to-end so the UI can render citations.
- Snapshot tests cover payload shape.

PR #99, opened 2026-05-05
feat(rag): hybrid retriever — semantic + lexical + link-graph (#8)

### LINKED ISSUE #8: Retrieval: hybrid SQL (HNSW + BM25 + link-graph boost)
Single-pass SQL that joins:

- HNSW top-k over `embeddings.vec` (cosine).
- BM25 hits from the FTS index on `chunks.text_for_embed`.
- One- and two-hop graph boosts from `links` (multiplicative, as in the published Obsidian RAG build).

Each result row is tagged with the winning signal: `semantic | lexical | backlink | tag | recency_boost`.

### Acceptance
- Query latency P50 ≤ 80 ms on a 5K-note vault (per planning/rag.md §"Steady state").
- Backlink-boosted results outrank pure-semantic matches on the golden-set tests that target link structure.
- Result rows always carry `winning_signal`.

heimdallSWE:
## What
Closes #8.

Lands the full hybrid-retrieval stack for RAG v1, mapped onto the TS-MVP path (the issue's original SQL/HNSW phrasing was written for the DuckDB stack that became archived/rust-only):

- **`InMemoryLinksIndex`** + **`LinksIndexService`** — forward and reverse map of intra-vault wikilinks and markdown links, kept in sync with the runner. O(edges-affected) link promotion via an `unresolvedByTarget` index, so cyclic and out-of-order arrivals work without rescanning. Self-links excluded from backlinks/neighborCount; rename rewrites resolved targets and re-runs promotion/demotion.
- **`InMemoryBM25Index`** + **`BM25IndexService`** — standard BM25 (k1=1.2, b=0.75), Unicode-letter tokenizer that preserves Swedish å/ä/ö, content-addressed by `contentHash` so a hash exists at most once. Same orphan-eviction semantics as `EmbeddingService` (via `IngestionStore.isHashReferenced`).
- **`HybridRetriever implements Retriever`** — RRF (k=60) over the union of top-K\*5 candidates from semantic + lexical, multiplicative link-graph boost capped at 2.0, deterministic tie-breaking on `(path, ord)`. Tags each row with `winningSignal` (`semantic | lexical | backlink`).
- **Contracts**: `RetrievalHit`, `RetrievalPayload`, `WinningSignal`, `Retriever`, `PayloadAssembler` — the typed boundary the chat-side issues (#10, #14) build against.
- **`IngestionStore.getChunksByHash`** — the contract addition needed to hydrate `RetrievalHit` rows from a hash-keyed search. Both `InMemoryIngestionStore` and `JsonIngestionStore` updated.
- **Golden-set seed** — `evals/golden/link-structure.jsonl` + a 10-note fixture micro-vault under `evals/golden/fixtures/link-graph/` exercising hub fan-out, backlink fan-in, and 2-hop noise.

## Why
- #8 is the foundation for the chat tool-loops (#13, #14) — they need a single typed retrieval entry point.
- RRF over BM25 + cosine is robust against the very different score scales each side produces; the multiplicative link boost matches the Obsidian RAG build that planning/rag.md references.
- The `RetrievalHit`/`RetrievalPayload` contracts are pinned now so #10 (payload assembler) and #14 (query loop) can be built independently.

## How to test
- `npm test` — 230/230 green.
- `npx vitest run src/services/hybrid-retriever.test.ts` — 12 unit tests with mocked dependencies.
- `npx vitest run src/services/hybrid-retriever.integration.test.ts` — 8 integration tests against the link-graph fixture vault.
- `npx tsc --noEmit` — clean.

## Notes for review
- **Boost-effect validation against ideal-paths is deferred to #16's eval harness.** The `MockEmbedder` used in tests is deterministic but semantically arbitrary, so per-query rank flips between with-and-without-`LinksIndex` are not reliable enough to assert in a unit-style test. The integration test verifies wiring + that link-graph affects orderings *somewhere*, not direction-of-effect on specific golden questions. Real-embedder validation lands with #16.
- Link tag rule is `linkExcess > 0 → "backlink"`. Rejected stricter rules where `linkExcess > sem` because the multiplicative cap of 2.0 makes that structurally impossible — see commit message on `aac7f94`.
- `winningSignal` union is forward-compatible with `tag` and `recency`; v1 only emits the first three.

## Issue scope vs TS-MVP path
| Acceptance | Status |
|---|---|
| Single-pass fused retrieval | ✅ TS equivalent: RRF over the three signals |
| Each result row tagged with winning signal | ✅ |
| Backlink-boosted results outrank pure-semantic on link-structure golden set | ⚠️ Plumbing + fixtures landed; rigorous validation deferred to #16 |
| P50 ≤ 80 ms on 5K-note vault | ⏳ Brute-force cosine + BM25 is structurally well under budget; verification needs #16's harness to time it |


SimonProjekt (changes requested):
Caught a build break before merging.

**Issue:** \`src/contracts/index.ts:13\` exports from \`./ingest-tool-loop\` but the file does not exist anywhere on this branch.

**Origin:** Introduced in commit cb52325a (the BM25 commit) alongside the \`./bm25-index\` export. Looks like the file was intended to land in the same commit but was missed.

**Effect:**

- \`npx tsc --noEmit\` → \`error TS2307: Cannot find module './ingest-tool-loop'\`
- \`npm test\` → 6 test files fail to load with the same error; only 160 of the ~230 tests actually run. The PR description's "230/230 green" likely reflects local state where the file existed as untracked.

**Fix:** Either drop the broken export line:

\`\`\`diff
 export * from \"./bm25-index\";
-export * from \"./ingest-tool-loop\";
\`\`\`

…or add the intended \`src/contracts/ingest-tool-loop.ts\` to the branch. A grep for \`ingest-tool-loop\` / \`IngestToolLoop\` across \`src/\` and \`evals/\` returns only this one barrel-export line, so dropping it should be safe.

Once that lands I'll re-review and merge.

heimdallSWE:
Fixed in f938927 — dropped the stale `export * from "./ingest-tool-loop"` line.

You diagnosed it exactly: the file was sitting in my working tree as untracked #13 work-in-progress from a parallel agent, so my local `tsc` and `vitest` resolved happily while CI saw the dangling export. After clearing those untracked files, I reproduced the CI failure locally, then applied the one-line fix.

Local now: `tsc --noEmit` clean, `npm test` 223/223 (the 7-test delta vs my earlier "230" was the untracked `ingest-tool-loop.test.ts` no longer being discovered — that's correct; it doesn't belong on this branch).

Ready for re-review.

SimonProjekt (approved):
Fix verified locally: `tsc --noEmit` clean, 213/213 tests pass. Spot-check on the new retrieval contract and HybridRetriever — well-documented strategy, extracted constants, deterministic tie-breaking, edge cases covered. Approving and merging.

PR #95, opened 2026-05-05
docs: add CONTRIBUTING.md with team git workflow

SimonProjekt:
## What
Adds CONTRIBUTING.md documenting the team's git workflow for Gemmera.

## Why
We are now three on the team and need clear shared rules for branch strategy, commit format, PR process, and review flow. The document is also intended to be saved as Claude memory so AI pair-programmers follow the same rules as the team.

## Contents
- Branch strategy: `main` / `dev` / `feature/...`
- Conventional Commits format with examples taken from this repo's history
- PR process: description template, review flow, squash & merge as default
- Conflict handling step-by-step
- List of destructive operations we never perform
- A dedicated section at the bottom aimed at Claude (AI pair)

## How to review
Read the document and comment if anything feels unclear, insufficient, or wrong. The rules are proposals — if something does not match how you want to work, say so before merge.

Note in particular:
- The branch naming convention (`feature/`, `fix/`, `docs/`, `refactor/`)
- Squash & merge as the default merge strategy
- That one PR equals one logical reviewable unit, not necessarily one issue
- The English-only rule for all repo and GitHub content (replaces the Swedish version closed in #94)

Issue #60, opened 2026-04-27
Tool-loop: cancellation + per-turn timeout with shared unwind


heimdallSWE:
Implement first-class cancellation and timeout that drive the shared unwind from #2.

### Acceptance
- User can abort at any state via a UI affordance; abort triggers `unwind("cancelled")` and the loop exits to `CANCELLED`.
- Per-turn wall-clock timer (default 120 s, configurable up to 300 s, sourced from the hard-stops constants in #4) fires `unwind("timed_out")` on expiry.
- Unwind order matches #2: stop model stream → drop pending tool results → roll back unconfirmed writes → write events → Notice → close turn.
- Confirmed writes from prior states are preserved across cancel/timeout.
- Tests cover: cancel during `GENERATE`, cancel during a long tool call, timeout during `RERANK`, timeout during `GENERATE`.

### Reference
- `planning/tool-loop.md` — "Cancellation support", "Timeout support".

heimdallSWE:
**Status update — partially landed in #100 (now in `dev`).**

**Done:**
- `sm.cancel(terminalState?)` transitions immediately to the cancellation terminal (default `CANCELLED`) via the shared unwind from #34
- Per-turn `AbortController` exposed on `StateContext.signal`; aborted as the first step of `runUnwind` and immediately on `cancel()` and wall-clock timeout
- `JsonConstrainedDecoder.decode` accepts and forwards `AbortSignal` to `LLMService.chat` (which already supports it); concrete states pass `StateContext.signal` so model streams release on terminal transitions
- Wall-clock timer with `budgetMs` validated to `(0, 300_000]` ms drives `TIMED_OUT` through the same unwind path
- Confirmed writes preserved across cancel/timeout via the consumer-side `rollbackUnconfirmedWrites` hook

**Remaining for this issue to close:**
- Concrete-state cancellation/timeout tests (cancel during `GENERATE`, cancel during a long tool call, timeout during `RERANK`, timeout during `GENERATE`) — these live with #39 / #41 once the concrete states exist
- UI cancel affordance — UI-surfaces v1

SimonProjekt:
I'll take this.

Issue #57, opened 2026-04-27
Tool-loop: intent classifier integration (skips, thresholds, disambiguation chip)


heimdallSWE:
Integrate the intent classifier into the front of the tool loop, with skip conditions, disambiguation chip, and the asymmetric thresholds. The classifier prompt itself is owned by #5; this issue owns the routing surface.

### Acceptance
- Every user message hits the classifier *unless* a skip condition fires: empty (error), attachment-only (`capture`), command-invoked, `Ctrl/Cmd+Enter` (`capture`), leading `?` (`ask`).
- Every skip writes a `classifier_decision` event with `source: "skip"` and the reason.
- Output JSON validated under `format: "json"` (uses #11). Schema-bad output → 1 retry → fall back to `ask` and show disambiguation chip.
- Asymmetric thresholds in config (anchors: `ask` 0.70, `capture` 0.85, `mixed` 0.75, `meta` 0.70). Below threshold → disambiguation chip with `[Save] [Ask] [Cancel]`.
- Edge cases: classifier timeout (>500 ms) → `ask` + chip; unparseable → `ask` + chip; out-of-taxonomy label → `ask` + chip; Ollama unreachable → main-loop Ollama-error handling.
- Disambiguation result writes a `classifier_disambiguation` event for golden-set growth.
- Classifier retries do not count against the main-turn 3-retry budget (#4).

### Reference
- `planning/classifier.md` — full spec.
- `planning/tool-loop.md` — "Intent classifier".

iamaxeloberg:
Implemented in commit 4414071 on feature/wire-ollama-llm-factory.

Issue #54, opened 2026-04-27
Tool-loop: JSON-schema constrained decoding at every structured-output state


heimdallSWE:
Wire JSON-schema-constrained decoding into every structured-output state. Per `rag.md`, this is the response of choice if Gemma's tool calls underperform; v1 ships with it on by default.

### Acceptance
- `PARSE_CONTENT`, `DECIDE_STRATEGY`, `VALIDATE_CITATIONS`, and the classifier all submit with Ollama `format: "json"` (or grammar) plus the relevant JSON schema.
- A schema registry maps state → schema. Schemas live alongside the prompt files for that state.
- Output that parses but violates schema feeds the validation-retry path (#10).
- A small benchmark utility exercises each constrained call against fixtures and reports schema-failure rate (a stand-in metric for the eval suite).
- Documented escape hatch: a single setting toggles constrained decoding off for debugging, but never in user-facing builds.

### Reference
- `planning/tool-loop.md` — "Retry policy" → Model-output validity.
- `planning/rag.md` — "If Gemma's tool calls underperform".

heimdallSWE:
**Status update — partially landed in #100 (now in `dev`).**

**Done:**
- `SchemaRegistry` contract + `InMemorySchemaRegistry`
- `JsonConstrainedDecoder` with ajv validation, validator cache keyed by schema identity, function-form prompt assembly, `AbortSignal` forwarding
- Discriminated `ConstrainedDecodeResult` (`reason: "parse" | "schema"`) so the retry router in #51 can branch
- `benchmarkDecoder` utility reporting `parseFailures`, `schemaFailures`, `successRate`
- Escape-hatch option that throws when `NODE_ENV === "production"` and warns otherwise
- Placeholder JSON schemas at `prompts/<id>.schema.json` so the file-layout contract is encoded
- `TODO(format-json)` in `composeSystemPrompt` flagging where to drop inline schema text once the LLM service grows native `format:"json"`

**Remaining for this issue to close:**
- Real v1 schemas (replace the permissive `{"type":"object"}` placeholders) for `PARSE_CONTENT`, `DECIDE_STRATEGY`, `VALIDATE_CITATIONS`, and the classifier
- Native Ollama `format:"json"` wiring once `LLMService.ChatOptions` grows that field
- Schemas wired into `PARSE_CONTENT`, `DECIDE_STRATEGY`, `VALIDATE_CITATIONS`, and the classifier (concrete states land in #39/#41)

PR #100, opened 2026-05-05
feat(tool-loop): state machine foundation, observability, reliability, prompts

### LINKED ISSUE #33: Tool-loop: state machine framework (named states, transitions, queueing)
Build the generic state-machine framework that both the ingest and query loops sit on top of. This is infrastructure only — no tool wiring yet. Pinned in `planning/tool-loop.md` under "State machine requirements" (structural section).

### Acceptance
- A `StateMachine` abstraction with: named states, single-responsibility per state, explicit named transitions, no implicit fall-through.
- One active turn per chat enforced; second submission is queued (returns a queued handle the UI can render as a chip).
- Bounded events per turn — every state declares a max event count; exceeding it transitions to a terminal error state.
- Transitions are event-driven only: `user_action`, `tool_result`, `timer`, `model_output`. No "tick" loops.
- Each state defines an `enter` and `exit` hook with payload, used downstream for logging.
- Unit tests for: queueing a second turn, bounded-event enforcement, illegal transition rejection.

### Reference
- `planning/tool-loop.md` — "State machine requirements" → Structural & Behavioral.

SimonProjekt:
## What

Lays the foundational framework for the Gemmera tool loop. Eight closely-coupled issues covering the state machine, terminal states with a shared unwind, event logging with deterministic replay, per-turn hard stops, the prompt library scaffold, the layered retry policy, JSON-schema constrained decoding, and the cancellation API.

## Why

Tool-loop v1 sits between the chat UI and the RAG layer. The ingest and query state machines (#39, #41) and the intent classifier integration (#57) all depend on this foundation. Landing the framework, observability, and reliability layers as one coherent bundle lets downstream issues wire concrete states without re-litigating framework design.

## Changes since first review

Six follow-up commits address @heimdallSWE's review:

1. **fix(tool-loop): isolate unwind hooks, synthesize bounded-events event** — runUnwind wraps each hook in safeRun so a thrown hook does not strand the rest, the timer, or the queue. The bounded-events ceiling now synthesizes its own `{kind: "limit", name: "bounded_events"}` event. Documents counter semantics and the timer/dispatch race.
2. **fix(tool-loop): replay counter- and retry-triggered turns; DuckDB TODO** — bumpCounter writes a `bump` log entry on every call so replay can reconstruct counter-triggered paths. replayTurn handles bump entries, framework-emitted limit/retry/timer events, and skips wall-clock-only events. Replay tests added for both counter-exhaustion and retry-exhaustion turns.
3. **feat(tool-loop): per-turn AbortSignal in StateContext** — StateMachine owns a fresh AbortController per turn; StateContext.signal exposes its signal. Aborted as the first step of runUnwind, and immediately on cancel() / wall-clock timeout. JsonConstrainedDecoder.decode forwards the signal to LLMService.chat.
4. **refactor(tool-loop): tighten API per review** — StateMachineEventKind splits into ConsumerEventKind (matched by transitions) and FrameworkEventKind (synthetic-only). Constructor validates timer.budgetMs is in (0, 300000] ms. Validator cache keyed by schema identity (WeakMap) so re-registering recompiles. TODO(format-json) note. Idempotency-test comment links to the future no-duplicate-effects assertion in #39.
5. **feat(tool-loop): canonical hard-stop constants and placeholder prompt schemas** — `src/contracts/hard-stops.ts` pins the per-turn ceilings from planning/tool-loop.md. `prompts/<id>.schema.json` placeholder schemas lock the file-layout contract for the six v1 prompts.
6. **merge: sync dev** — resolves the `src/contracts/index.ts` conflict with the retrieval exports from #99 and the contributing guide from #95.

Deferred per the review's "nice-to-haves I won't block on": typed payload union for `triggeringEvent.payload` (waits for concrete state machines to inform the shape) and YAML frontmatter for prompt files.

## What is in this PR

| Issue | Summary | Status |
|---|---|---|
| #33 | StateMachine framework — named states, explicit transitions, per-turn queueing, bounded events, enter/exit hooks | Closes |
| #34 | Canonical terminal states and the shared unwind order | Closes |
| #35 | EventLog contract, InMemoryEventLog, transition-by-transition logging, replayTurn helper, redactEvent hook | Refs (DuckDB writer pending) |
| #36 | Per-turn hard stops — wall-clock timer, named per-turn limits, canonical constants | Refs (ceiling-to-terminal mappings and 16 KB payload primitive land with concrete states) |
| #37 | prompts/ directory with six versioned Markdown stubs and schema placeholders, FilePromptLoader | Refs (cowork.version stamping waits on the note writer) |
| #51 | Retry primitive — per-state retry budget integrating with the global per-turn cap | Refs (tool-call and transient-infra retry layers wire with concrete states) |
| #54 | Constrained decoding — SchemaRegistry, JsonConstrainedDecoder with ajv validation, benchmark utility, escape-hatch option | Refs (v1 schemas with real fields land with the prompts they constrain) |
| #60 | sm.cancel() through the shared unwind, AbortSignal propagation | Refs (concrete-state cancellation/timeout tests live with #39 and #41) |

## How to test

- `npm test` — 272 tests across 29 files pass
- `npx tsc --noEmit` — clean
- The framework surface lives in [src/contracts/state-machine.ts](src/contracts/state-machine.ts) and [src/services/state-machine.ts](src/services/state-machine.ts); usage examples are the unit tests under [src/services/](src/services/).

## What is not in this PR

Deferred to downstream issues that own those flows:

- Concrete state names like PARSE_CONTENT, RETRIEVE — #39, #41
- Native Ollama `format: "json"` wiring — requires extending LLMService.ChatOptions
- `cowork.version` stamping into notes and the events log — depends on the note writer (#52, #53) and events.duckdb
- DuckDB-backed event log writer — depends on events.duckdb
- UI integration of the cancel button and status reflection — UI-surfaces v1

## Dependency change

Adds `ajv ^8.20` as a runtime dependency for JSON Schema validation.

Closes #33, #34

Refs #35, #36, #37, #51, #54, #60


src/services/state-machine.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  private async runUnwind(ctx: StateContext): Promise<void> {
+    const u = this.config.unwind;
+    if (!u) return;
+    await u.stopModelStream?.(ctx);
+    await u.dropPendingToolResults?.(ctx);
+    await u.rollbackUnconfirmedWrites?.(ctx);
+    await u.writeEvents?.(ctx);
+    await u.surfaceNotice?.(ctx);
+  }
```
heimdallSWE, line 361: **runUnwind error isolation.** If any hook throws, the remaining hooks don't run, `clearTimer()` and `this.active = null` below are skipped, and the queue stalls forever. Wrap each call in a try/catch (logging the error) so unwind is best-effort. This is the biggest real defect in the PR.

src/services/state-machine.ts
```diff
@@ -0,0 +1,299 @@
...
+    const fromState = this.active.currentState;
+    const fromDef = this.states.get(fromState)!;
+    if (fromDef.terminal) {
+      throw new Error(`Cannot dispatch from terminal state: ${fromState}`);
+    }
+
+    const newCount = (this.eventCounts.get(fromState) ?? 0) + 1;
+    this.eventCounts.set(fromState, newCount);
+    if (newCount > fromDef.maxEventsPerTurn) {
+      await this.transitionTo(this.config.errorBoundedEventsState, event);
```
heimdallSWE: **Bounded-events terminal carries the wrong event.** When the bounded-events ceiling is exceeded, we forward the *user's* event into the terminal. The consumer can't distinguish "we hit the bounded-events cap" from any normal transition. `bumpCounter` already synthesizes `{kind: "limit", name}` — recommend doing the same here, e.g. `{kind: "limit", name: "bounded_events"}`.

src/services/state-machine.ts
```diff
@@ -0,0 +1,299 @@
...
+      fromState: null,
+      triggeringEvent: null,
+    };
+    await this.writeLog("enter", ctx);
+    await this.states.get(initial)!.onEnter?.(ctx);
+    if (this.config.timer) {
+      const { budgetMs, terminalState } = this.config.timer;
+      this.timeoutId = setTimeout(() => {
+        void this.handleTimeout(terminalState);
+      }, budgetMs);
```
heimdallSWE, line 254: **No default/clamp on `budgetMs`.** Spec says default 120 s, configurable up to 300 s. Framework accepts any positive number. Worth defaulting + clamping (or at least asserting) so a misconfigured 600 s doesn't slip through.

src/services/state-machine.ts
```diff
@@ -0,0 +1,299 @@
...
+    const next = (this.counterValues.get(name) ?? 0) + 1;
+    this.counterValues.set(name, next);
+    if (next > limit.limit) {
+      await this.transitionTo(limit.terminalState, {
+        kind: "limit",
+        name,
+        payload: { limit: limit.limit, value: next },
+      });
+    }
+  }
```
heimdallSWE, line 168: **Counter semantics undocumented.** `if (next > limit.limit)` means "the limit value is the max allowed; the bump that pushes past it triggers the terminal." Tests pin this, but a one-line doc comment would prevent confusion when teammates wire `MAX_TOOL_CALLS = 10` and bump once per call.

src/services/state-machine.ts
```diff
@@ -0,0 +1,299 @@
...
+      state: initial,
+      fromState: null,
+      triggeringEvent: null,
+    };
+    await this.writeLog("enter", ctx);
+    await this.states.get(initial)!.onEnter?.(ctx);
+    if (this.config.timer) {
+      const { budgetMs, terminalState } = this.config.timer;
+      this.timeoutId = setTimeout(() => {
+        void this.handleTimeout(terminalState);
```
heimdallSWE, line 253: **Timer/dispatch race.** The `setTimeout` callback fires `void this.handleTimeout(...)` with no async lock against an in-flight `dispatch()`. Two paths can both call `transitionTo` and race on `this.active`. Mostly latent given how state machines will be driven, but worth either a comment acknowledging it or a small "transition lock" / re-check in `handleTimeout`.

src/contracts/state-machine.ts
```diff
@@ -0,0 +1,128 @@
+export type StateMachineEventKind =
+  | "user_action"
+  | "tool_result"
+  | "timer"
+  | "model_output"
+  | "limit"
+  | "retry";
```
heimdallSWE: **Event kind union mixes consumer and framework kinds.** `limit` and `retry` are framework-internal but appear in the same union as `user_action`/`tool_result`/etc. Consumers can write transitions on framework kinds that will never fire. Suggest splitting:
```ts
type ConsumerEventKind = "user_action" | "tool_result" | "timer" | "model_output";
type FrameworkEventKind = "limit" | "retry";
```

src/contracts/state-machine.ts
```diff
@@ -0,0 +1,128 @@
...
+// output the consumer attaches as `payload`). Used by the turn inspector
+// and as the input to deterministic replay.
+export interface EventLogEntry {
+  turnId: string;
+  kind: EventLogEntryKind;
+  state: string;
+  fromState: string | null;
+  timestamp: number;
+  triggeringEvent: StateMachineEvent | null;
+}
```
heimdallSWE, line 135: **Typed payload would help #39/#41.** `triggeringEvent.payload: unknown` is the only carrier for "tool args, tool results, prompts, model responses" mentioned in the planning doc. Concrete state machines will each have to invent their own payload conventions. Even a documented union (`ToolArgsPayload | ToolResultPayload | ...`) here would prevent drift.

src/services/event-log.ts
```diff
@@ -0,0 +1,41 @@
...
+    .map((e) => e.triggeringEvent!);
+
+  const replayLog = new InMemoryEventLog();
+  const sm = new StateMachine({ ...config, eventLog: replayLog });
+  await sm.startTurn(turnId);
+  for (const event of dispatchSequence) {
+    await sm.dispatch(event);
+  }
+  return replayLog.eventsFor(turnId);
+}
```
heimdallSWE, line 67: **Replay reconstruction is incomplete.** Filtering `enter` entries with non-null `triggeringEvent` and dispatching them works for event-driven turns, but breaks for turns with `bumpCounter` (kind `"limit"`), wall-clock timeouts (kind `"timer"`, name `"turn_timeout"`), or `retry()` (kind `"retry"`). Those entries are non-null, get re-dispatched, and `dispatch()` throws "No transition from X on limit:Y". Add a test that captures a counter-triggered turn and replays it — it should fail today.

src/services/event-log.ts
```diff
@@ -0,0 +1,41 @@
...
+  private entries: EventLogEntry[] = [];
+
+  write(entry: EventLogEntry): void {
+    this.entries.push(entry);
+  }
+
+  async eventsFor(turnId: string): Promise<readonly EventLogEntry[]> {
+    return this.entries.filter((e) => e.turnId === turnId);
+  }
+}
```
heimdallSWE, line 21: **No DuckDB writer.** Acknowledged in the PR description, but worth a TODO comment here so it's not invisible. #35's acceptance lists `events.duckdb` explicitly.

src/services/json-constrained-decoder.ts
```diff
@@ -0,0 +1,134 @@
...
+    prompt: string;
+  }): Promise<ConstrainedDecodeResult> {
+    const schema = this.registry.get(opts.stateName);
+    const response = await this.llm.chat({
+      messages: [
+        { role: "system", content: composeSystemPrompt(schema) },
+        { role: "user", content: opts.prompt },
+      ],
+    });
+    const raw = response.content;
```
heimdallSWE, line 59: **No `AbortSignal` forwarded.** When we wire cancellation properly, `decode()` should accept a signal and pass it to `chat()` (which already supports it via `LLMService.ChatOptions.signal`). Adding `signal?: AbortSignal` to the `decode` opts now is essentially free.

src/services/json-constrained-decoder.ts
```diff
@@ -0,0 +1,134 @@
...
+  private getValidator(
+    stateName: string,
+    schema: JsonSchema,
+  ): ValidateFunction {
+    const cached = this.validators.get(stateName);
+    if (cached) return cached;
+    const validator = this.ajv.compile(schema);
+    this.validators.set(stateName, validator);
+    return validator;
+  }
```
heimdallSWE, line 94: **Validator cache keyed by stateName.** If a consumer (or a test) re-registers a different schema for the same stateName, the cached validator goes stale. Either invalidate on re-register in the registry, or key the cache by schema identity.

src/services/json-constrained-decoder.ts
```diff
@@ -0,0 +1,134 @@
...
+function composeSystemPrompt(schema: JsonSchema | undefined): string {
+  if (!schema) {
+    return "Respond with valid JSON only. No preamble, no commentary, no code fences.";
+  }
+  return [
+    "Respond with valid JSON only. No preamble, no commentary, no code fences.",
+    "The JSON must conform to this schema:",
+    JSON.stringify(schema, null, 2),
+  ].join("\n\n");
+}
```
heimdallSWE, line 110: **Schema inlined in every prompt.** Inlining `JSON.stringify(schema, null, 2)` into the system prompt bloats every call. Once `LLMService.ChatOptions` grows native `format: "json"`, this can drop to a one-liner instruction. A `// TODO(format-json)` near `composeSystemPrompt` would mark the cleanup.

src/services/schema-registry.ts
```diff
@@ -0,0 +1,27 @@
...
+    return this.schemas.has(stateName);
+  }
+
+  entries(): readonly { stateName: string; schema: JsonSchema }[] {
+    return Array.from(this.schemas.entries()).map(([stateName, schema]) => ({
+      stateName,
+      schema,
+    }));
+  }
+}
```
heimdallSWE, line 27: **No v1 schemas committed.** #54 acceptance: "Schemas live alongside the prompt files for that state." We ship the registry but no actual schemas. Suggest committing minimal placeholder JSON schemas at `prompts/<id>.schema.json` (even `{"type":"object"}`) so the file-layout contract is encoded and concrete states can `register()` against committed files rather than ad-hoc objects.

src/services/file-prompt-loader.ts
```diff
@@ -0,0 +1,47 @@
...
+
+function parsePromptFile(raw: string): { version: string; body: string } {
+  const match = raw.match(/^version:\s*(\S+)\s*\n([\s\S]*)$/);
+  if (!match) {
+    throw new Error(
+      'Prompt file must start with a "version: <semver>" line',
+    );
+  }
+  return { version: match[1], body: match[2].trim() };
+}
```
heimdallSWE, line 47: **Single-line `version:` instead of YAML frontmatter.** Cheap to switch now to standard YAML frontmatter (`---\nversion: 0.1.0\n---`). Lets us add `model`, `temperature`, etc. later without changing the parser. Painful to migrate after teammates start writing prompts against this format.

src/services/state-machine.test.ts
```diff
@@ -0,0 +1,651 @@
...
+      errorBoundedEventsState: "ERROR_BOUNDED",
+    };
+    const sm = new StateMachine(config);
+    await sm.startTurn("turn-1");
+    await sm.dispatch({ kind: "user_action", name: "review" });
+    await sm.dispatch({ kind: "user_action", name: "rewrite" });
+
+    expect(onEnterCount).toBe(2);
+    expect(sideEffectCount).toBe(1);
+  });
```
heimdallSWE, line 291: **Idempotency test asserts the opposite of #34.** The docstring `"consumer is responsible for idempotency"` flips #34's acceptance ("Tests assert retried states do not duplicate effects"). Reasonable deferral since `WRITE` doesn't exist yet, but worth a comment linking to the future test in #39/#41 so the gap doesn't get lost.

heimdallSWE (changes requested):
**Overall:** Strong framework PR. The mechanisms are clean, well-typed, and the framework-level test coverage is genuinely thorough (50 new tests across 5 files, all passing locally after `npm install`). My concern is scope tracking, not code quality: several of the eight referenced issues have acceptance criteria that aren't met by mechanism alone, and the PR auto-closes them all. I'd like us to (a) fix a small set of real defects, (b) weaken some `Closes #X` to `Refs #X` so we don't lose the remaining work, and (c) ship a few constants/scaffolds that are cheap now and painful to retrofit later.

**Defects to fix before merge**

1. `runUnwind` is not error-isolated — if any hook throws, the rest of unwind doesn't run, the timer isn't cleared, `this.active` isn't nulled, and the queue stalls. Wrap each hook in try/catch.
2. `replayTurn` is unsound for any turn that hit a counter limit, the wall-clock timer, or `retry()` — those entries carry framework-internal triggering events (`kind: "limit" | "timer" | "retry"`) that have no defined transitions, so replay throws. Either filter them out and re-invoke `bumpCounter`/`retry` from a parallel recorded stream, or widen `dispatch()` to handle them.
3. `sm.cancel()` doesn't propagate an `AbortSignal` to in-flight LLM calls. `LLMService.ChatOptions.signal` already exists; we just don't wire it. Suggest the `StateMachine` own a per-turn `AbortController`, expose `ctx.signal` on `StateContext`, and abort it as the first step of `runUnwind`.

**Cheap-now, painful-later additions**

4. Add `src/contracts/hard-stops.ts` with the canonical numbers from `planning/tool-loop.md` (10 tool calls, 120 000 ms, 3 no-ops, 3 retries, 16 384 bytes). Even unwired, the constants shouldn't drift across teammates.
5. Ship placeholder JSON schemas next to the prompts (e.g. `prompts/ingest-parser.schema.json`) so the registered-schema contract is encoded — even an empty `{ "type": "object" }` is enough to lock the file layout.

**Scope-tracking changes**

Suggest weakening these `Closes #X` to `Refs #X` so the remaining acceptance bullets stay tracked:

| Issue | Why keep open |
|---|---|
| #35 | DuckDB writer not wired; replay soundness gap |
| #36 | Canonical constants + ceiling→terminal mappings + 16 KB payload primitive missing |
| #37 | `cowork.version` stamping not done (acknowledged) |
| #51 | Entire policy layer not implemented — only the retry primitive |
| #54 | v1 schemas not committed alongside prompt files |
| #60 | `AbortSignal` not propagated; concrete-state tests deferred |

Keep `Closes` for **#33** and **#34**, which are genuinely complete.

Nice-to-haves I won't block on: split `StateMachineEventKind` into consumer vs framework unions, switch prompt files to YAML frontmatter, add typed payload schema for event-log entries, gate `escapeHatch` behind `NODE_ENV !== "production"`.

iamaxeloberg (commented):
Seconding @heimdallSWE's review — the three blocking defects and scope-tracking changes are spot-on. A few additional things I noticed that aren't covered:

**Validation gaps in the constructor**

- `errorBoundedEventsState` is validated to exist but never checked for `terminal: true`. If someone wires a non-terminal state there, bounded-events protection silently breaks.
- A state with both `retry` config and `terminal: true` passes validation, but `retry()` is unreachable there (active is nulled on entry). Should reject at config time.

**`InMemoryEventLog.eventsFor` is a false async**

It returns synchronously but the `EventLog` contract types it as `Promise<...>`. Works now (auto-wraps), but a future DuckDB-backed log assuming `eventsFor` implies settled writes will race.

**`transitionTo` has no active-turn guard**

Every caller checks `this.active` beforehand, but `transitionTo` itself uses `this.active!` with a non-null assertion. A defensive guard at the top would prevent a future code path from panicking at runtime.

**`escapeHatch` not gated**

The docstring says "never enable in user-facing builds" but the `JsonConstrainedDecoder` option has no enforcement. At minimum log a warning, ideally check `NODE_ENV !== "production"`.

All non-blocking observations — the framework is otherwise solid and the test coverage is excellent.

SimonProjekt:
Thanks for the thorough review — addressed in 6 follow-up commits. Summary in the updated PR description under "Changes since first review".

**Critical defects:** all three fixed (runUnwind error isolation, replay reconstruction for counter-/retry-triggered turns, AbortSignal propagation through StateContext + decoder).

**Inline issues:** all addressed except the two you flagged as nice-to-haves — typed payload union for triggeringEvent.payload (waits for concrete state machines to inform the shape) and YAML frontmatter for prompt files.

**Scope tracking:** weakened Closes → Refs for #35, #36, #37, #51, #54, #60 per your table. #33 and #34 stay as Closes.

**Test count:** 272/272 (29 files, includes the work merged from dev — #95, #96, #99). Two new tests for the bumpCounter/retry replay paths, one for the AbortSignal propagation, and a couple of small additions for the bounded-events synthetic event and the unwind-isolation invariant.

Ready for re-review whenever you have a chance.

SimonProjekt:
@iamaxeloberg thanks — all four observations addressed in commit ee03480.

- **Constructor validation**: errorBoundedEventsState must now be marked terminal; states cannot have both retry config and terminal: true. Both rejected at config time with clear errors. Tests added.
- **EventLog write/read ordering**: documented the contract guarantee that any entry passed to a fully-resolved write is visible to a subsequent eventsFor for the same turn. The InMemoryEventLog already satisfies this; the contract pins the rule for the DuckDB-backed implementation.
- **transitionTo active-turn guard**: explicit check at the top of the method, captured into a local non-null variable so the rest of the body drops the `!` assertions.
- **escapeHatch gating**: now warns at construction when enabled, and throws when `NODE_ENV === "production"`. Defensive process.env access keeps it working in non-Node environments.

274/274 tests pass.

heimdallSWE (approved):
All three defects and the cheap-now items from the previous review are addressed and verified in code. Replay is sound on counter/retry paths, cancellation propagates AbortSignal to in-flight LLM calls, runUnwind is error-isolated, hard-stop constants and placeholder schemas are in place, and the PR body now correctly tracks scope (Closes #33, #34; Refs #35, #36, #37, #51, #54, #60). 274/274 tests pass, tsc clean. LGTM.

Issue #81, opened 2026-04-27
Classifier UI: disambiguation chip in composer header


heimdallSWE:
Implement the disambiguation chip in the composer header. Triggered when the classifier decision is below threshold OR on timeout/parse fallback.

Behavior:
- Renders within 250 ms of the classifier returning or timing out.
- Buttons: Save, Ask, Cancel. Tooltip on the chip shows the one-line rationale.
- Save / Ask re-submit the original message with the chosen intent hard-coded (same path as a command skip).
- Cancel discards the message.
- Non-modal: user can scroll, open files, change panes while it is visible.
- If the user types a new message while the chip is showing, that message is queued; the chip still applies to the first message only.

Emit a `classifier_disambiguation` event on any of the three button actions.

### Acceptance
- Visual + interaction tests cover all three buttons plus the queued-message case.
- 250 ms budget is met on a warm load (asserted in a perf test).
- Non-modal behavior verified by automated UI test.

### Reference
planning/classifier.md — "Disambiguation UX".

PR #109, opened 2026-05-06
fix(classifier): preserve null original_label for fallback disambiguations

heimdallSWE:
## Summary

Follow-up to #108. The disambiguation chip fires both for low-confidence classifications **and** for classifier fallbacks (timeout / unparseable / invalid-label / invalid-confidence). For the latter, \`decision.output\` is null, so on Save / Ask / Cancel we were writing rows like:

\`\`\`ts
originalLabel: cancelled.originalDecision.output?.label ?? "ask",
originalConfidence: cancelled.originalDecision.output?.confidence ?? 0,
\`\`\`

That stamps every fallback row as if the model had confidently emitted \`ask\` and the user corrected it. \`planning/classifier.md\` calls \`classifier_disambiguation\` "a built-in labeled-example factory — every user correction becomes a labeled example," so the golden set picks up mislabelled examples for the entire fallback population.

## Changes

- \`ClassifierDisambiguationRow.original_label\` → \`IntentLabel | null\`
- \`ClassifierDisambiguationRow.original_confidence\` → \`number | null\`
- DDL: drop \`NOT NULL\` on both columns (\`chosen_label\` was already nullable)
- \`toDisambiguationRow\` accepts the nullable types directly
- \`view.ts\` writes \`null\` for both fields when the original decision had no output, instead of defaulting to \`"ask"\`/\`0\`

The fallback reason is already on \`classifier_decision.fallback_reason\` and joinable via \`turn_id\`, so we don't need a new column to distinguish "model said X → user corrected" from "no model output → user picked something." Eval queries that count model-vs-user corrections should filter \`original_label IS NOT NULL\`.

## Tests

- \`toDisambiguationRow\` preserves null \`originalLabel\` / \`originalConfidence\`
- DDL guard against accidental re-tightening of \`NOT NULL\` on those columns

## Verification

- \`npm test\` — 422 tests pass (was 420 before this branch)
- \`npx tsc --noEmit\` — clean

Refs #81, #108.

### DIFF
--- a/src/contracts/classifier.ts
+++ b/src/contracts/classifier.ts
@@ -154,12 +154,19 @@ export interface ClassifierDecisionRow {
 /**
  * Row written to the `classifier_disambiguation` table when the user
  * corrects a low-confidence classification via the disambiguation chip.
+ *
+ * `original_label` and `original_confidence` are nullable because the chip
+ * also fires on fallback decisions (timeout / unparseable / invalid-label /
+ * invalid-confidence) where the model emitted no usable output. For those
+ * rows the original-side is null; the fallback reason can be joined from
+ * `classifier_decision.fallback_reason` via `turn_id`. Eval queries that
+ * count "model-vs-user" corrections should filter `original_label IS NOT NULL`.
  */
 export interface ClassifierDisambiguationRow {
   turn_id: string;
   ts: number;
-  original_label: IntentLabel;
-  original_confidence: number;
+  original_label: IntentLabel | null;
+  original_confidence: number | null;
   /** The label the user chose, or null when they cancelled. */
   chosen_label: IntentLabel | null;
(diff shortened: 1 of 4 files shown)


iamaxeloberg (approved):
Correct fix — the `?? "ask"` / `?? 0` defaults in #108 would have corrupted the golden set for the entire fallback population. Joining via `turn_id` → `fallback_reason` to distinguish model-vs-fallback rows is the right approach.

One thing to track before beta: the DDL uses `CREATE TABLE IF NOT EXISTS` so this schema change won't migrate any existing table. Fine pre-release, but needs an `ALTER COLUMN DROP NOT NULL` in whatever migration runner lands later — worth a follow-up issue or TODO in the DDL block.

Minor cosmetic: the two `view.ts` call sites are now identical and could share a helper, but not worth blocking on. Good to merge.

PR #111, opened 2026-05-06
feat(rag): wire retrieval stack and add payload assembler (#8, #10)

### LINKED ISSUE #8: Retrieval: hybrid SQL (HNSW + BM25 + link-graph boost)

### LINKED ISSUE #10: Retrieval: structured payload assembler for Gemma context

heimdallSWE:
## What

Two RAG v1 issues in one PR:

- **Wires the retrieval stack into the plugin lifecycle** (was previously orphaned). `services/index.ts` now constructs `InMemoryBM25Index` + `BM25IndexService`, `InMemoryLinksIndex` + `LinksIndexService`, and `HybridRetriever`, and exposes them on the `Services` bag. `main.ts` starts/stops the two new services next to `EmbeddingService` and wires matching debug logs.
- **Implements `DefaultPayloadAssembler` (#10)** per the contract in `src/contracts/retrieval.ts`: head-slices to `maxChunks` (default 8), projects `RetrievalHit → PayloadChunk` dropping retrieval internals, renames `winningSignal → whyMatched`, and resolves 1-hop neighbors via `LinksIndex` with per-chunk memoization and a configurable cap.

Side note: also rolls in the `.gitignore` tweak that was sitting uncommitted on `dev` — broadens the `demo-vault/**/.obsidian/` ignore globs so both `demo-vault/` and `demo-vault/raw/` work as vault roots.

## Why

`HybridRetriever` (PR #99, issue #8) was merged but never instantiated — `grep` for it in `main.ts` / `services/index.ts` returned nothing. Same for `BM25IndexService` and `LinksIndexService`. This PR un-orphans ~600 lines of tested code and unblocks #14 (query tool loop) by giving it a `retriever` and `payloadAssembler` to call.

#10 was already designed at the contract layer; this PR ships the implementation, leaving the chat-view integration as a clean follow-up.

## How to test

- `npm test` — 432 tests (was 420), 12 new for the payload assembler covering empty input, head-slice without re-sort, default `maxChunks=8`, projection + signal rename, `includeScores` flag, `includeNeighbors=false` field-omission, empty-array vs absent semantics, outgoing+backlinks dedup with self-skip, `maxNeighborsPerChunk` cap, neighbor memoization, and a full snapshot of a two-chunk payload with neighbors enabled
- `npx tsc --noEmit` — clean
- Manual: rebuild the plugin, reload Obsidian — debug logs should now show `[gemmera] bm25:added <path> (N)` and `[gemmera] links:indexed <path> (N)` after edits

Closes #8
Closes #10

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,12 +17,14 @@ styles.css.map
 .obsidian/workspace.json
 .obsidian/workspace-mobile.json
 
-# Obsidian vault runtime state
-demo-vault/raw/.obsidian/workspace.json
-demo-vault/raw/.obsidian/workspace-mobile.json
-demo-vault/raw/.obsidian/appearance.json
-demo-vault/raw/.obsidian/graph.json
-demo-vault/raw/.obsidian/plugins/
+# Obsidian vault runtime state — same set whether the vault root is
+# demo-vault/ or demo-vault/raw/.
+demo-vault/**/.obsidian/workspace.json
+demo-vault/**/.obsidian/workspace-mobile.json
+demo-vault/**/.obsidian/appearance.json
+demo-vault/**/.obsidian/graph.json
+demo-vault/**/.obsidian/plugins/
+demo-vault/.obsidian/
 
 # Dev
 .hotreload
--- a/src/main.ts
+++ b/src/main.ts
@@ -14,6 +14,8 @@ export default class GemmeraPlugin extends Plugin {
     this.services.eventBridge.start();
     this.services.ingestionRunner.start();
     this.services.embeddingService.start();
+    this.services.bm25IndexService.start();
+    this.services.linksIndexService.start();
     this.wireDebugLogs();
     // Fire reconcile in the background — hash gate keeps it cheap on warm reloads.
     this.services.reconciler
@@ -38,6 +40,8 @@ export default class GemmeraPlugin extends Plugin {
     if (this.services) {
       await this.services.ingestionRunner.stop();
       await this.services.embeddingService.stop();
+      await this.services.bm25IndexService.stop();
+      await this.services.linksIndexService.stop();
     }
     this.app.workspace.detachLeavesOfType(VIEW_TYPE);
   }
@@ -62,6 +66,26 @@ export default class GemmeraPlugin extends Plugin {
         return;
       }
       console.debug(`[gemmera] embed:${e.kind} ${e.path} (${e.count})`);
+    });
+    this.services.bm25IndexService.onEvent((e) => {
+      if (e.kind === "error") {
+        console.error("[gemmera] bm25 error", e);
+        return;
+      }
+      const count = "count" in e ? e.count : 0;
+      console.debug(`[gemmera] bm25:${e.kind} ${e.path} (${count})`);
+    });
+    this.services.linksIndexService.onEvent((e) => {
+      if (e.kind === "error") {
+        console.error("[gemmera] links error", e);
+        return;
+      }
+      if (e.kind === "renamed") {
+        console.debug(`[gemmera] links:renamed ${e.from}→${e.to}`);
+        return;
+      }
+      const count = "linkCount" in e ? e.linkCount : 0;
+      console.debug(`[gemmera] links:${e.kind} ${e.path} (${count})`);
     });
   }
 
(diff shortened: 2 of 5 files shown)


iamaxeloberg (commented):
**Overall: good PR, two things to flag.**

---

**`payload-assembler.ts` — `resolveNeighbors` doesn't break early**

The cap is applied after exhausting both the outgoing and backlinks loops. A hub note with 500 outgoing + 500 backlinks scans all 1000 before slicing to `maxNeighborsPerChunk` (default 5). Probably fine for v1, but cheap to fix: break out of each loop once `paths.length >= cap`.

---

**`Services` exposes `MutableLinksIndex`**

```ts
linksIndex: MutableLinksIndex;
```

Any consumer of the `Services` bag can now mutate the index directly, bypassing `LinksIndexService`. The bag should expose the read-only `LinksIndex` interface — `LinksIndexService` holds its own concrete reference internally and doesn't need to surface mutation capability upstream. Same question for `bm25Index: BM25Index` if that interface includes write methods.

---

**Minor notes:**
- Memoization test only spies on `outgoing`, not `backlinks` — the cache covers both, so not a correctness gap, just slightly incomplete coverage.
- Sequential `await stop()` in `onunload` is consistent with existing pattern; `Promise.all` would be faster at shutdown if that ever matters.
- `titleOf` uses basename over the stored `title` field — intentional per spec, but worth a comment for when frontmatter titles differ from filenames.

iamaxeloberg (commented):
Two observations:

## `payload-assembler.ts` — `resolveNeighbors` doesn't break early

```ts
for (const link of links.outgoing(path)) {
  if (!link.resolved || !link.target) continue;
  if (link.target === path) continue;
  if (seen.has(link.target)) continue;
  seen.add(link.target);
  paths.push(link.target);
}
// same loop for backlinks
// THEN: paths.slice(0, cap)
```

The cap is applied after exhausting both loops. A hub note with 500 outgoing + 500 backlinks scans all 1000 before slicing to `maxNeighborsPerChunk` (default 5). For v1 this is likely fine, but cheap to fix: `break` once `paths.length >= cap` in each loop.

---

## `Services` exposes `MutableLinksIndex`

```ts
linksIndex: MutableLinksIndex;
```

`MutableLinksIndex` presumably has `upsert`/`delete` on it, so any consumer of `Services` can mutate the index directly, bypassing `LinksIndexService`. The bag should expose the read-only `LinksIndex` interface instead — `LinksIndexService` (and anyone who needs to write) can hold its own reference to the concrete `InMemoryLinksIndex`. Same argument applies to `bm25Index` if that interface is also mutable.

heimdallSWE:
Thanks for the review — addressed in aef5d2a.

**`resolveNeighbors` early-break:** done. Added `if (paths.length >= cap) break;` to both loops. The post-loop `slice(0, cap)` is now redundant so I dropped it; the loops handle the bound directly.

**`Services` exposing `MutableLinksIndex`:** stronger fix than read-only typing — I dropped `bm25Index` and `linksIndex` from the `Services` interface entirely. `grep` confirmed nobody reads them off the bag (the retriever, assembler, and both index services already hold their own references via constructor injection), so the fields were dead surface area. This sidesteps the encapsulation question without needing a `Pick<>` or a new `ReadOnlyBM25Index` split.

**Minor notes:**
- Memoization test: now spies on `backlinks` too and asserts both counters equal 1. Full coverage of the per-path cache.
- Sequential `await stop()`: left as-is, matches the existing pattern.
- `titleOf` basename: added a comment explaining the consistency with `HybridRetriever.titleOf` and when it'd be worth revisiting (universal frontmatter title parsing).

432 tests still green.

PR #112, opened 2026-05-06
feat(rag): indexer ops surface (#15) + ingest tool loop (#13)

### LINKED ISSUE #13: Tools: ingest tool loop (with near-duplicate branch)

### LINKED ISSUE #15: Ops: cold-start UX, indexer controls, weekly reconciliation
Ship the operability surface described in planning/rag.md "Cold start vs steady state":

- "Indexing X of Y" pill in the chat header during cold start.
- "Pause indexer" toggle.
- Idempotent, resumable "Rebuild index" affordance in settings.
- Weekly reconciliation job that re-hashes every file and reconciles drift between the vault and `notes`.

### Acceptance
- Pausing the indexer halts new job claims within one tick.
- Rebuild is safe to interrupt and resume without duplicate rows.
- Reconciliation logs (and exposes) any divergence it finds.

heimdallSWE:
Closes #15
Closes #13

## Summary

- **#15 — Ops:** indexer status observable, pause/resume controls (persisted), epoch-based rebuild, weekly drift reconciliation, chat-header pill, settings tab.
- **#13 — Ingest tool loop:** parser → search_similar → decide_strategy → preview → write → update_index, with the near-duplicate branch.
- Self-audit pass: fixed RunnerStatus drain race, exact-hash dedup key, chip "Save" routing, append target drift, event-log emission, plus optimizations.
- Review-feedback fixes from #112 review: replaced rebuild monkey-patch with a typed constructor hook, looser heading match, byte-vs-semantic dedup comment, explicit confirm-on-dedup_ask guard.

## #15 (commit 16efc84)
- RunnerStatus aggregates pending/total/paused via runner.workSize().
- RunnerControls: pause persists in store meta, rebuild bumps a global epoch.
- IngestionRunner.setPaused halts new claims; in-flight finishes.
- HashGatedIngestionPipeline reads the rebuild epoch and re-stamps stale notes.
- ScheduledReconciler produces a DriftReport, runs immediately if overdue.
- IndexingPill widget + GemmeraSettingsTab (pause, rebuild w/ confirm, reconcile-now, drift summary).

## #13 (commit 60a5949)
- IngestWriter: atomic writeNew (cowork frontmatter + dated filename), appendUnderDatedHeading with mtime drift detection.
- parseContent: constrained-JSON LLM call to NoteSpec.
- decideStrategy: exact-hash dedup → threshold dedup_ask → LLM fallback → low-score short-circuit.
- runIngest: UI-agnostic orchestrator; emits one event-log row per phase.
- IngestPreviewModal (save/append/dedup); chat routes capture turns through it including from the disambiguation chip.
- Settings: inboxFolder, dedupThreshold, alwaysPreviewBeforeSave.

## Self-audit fixes (commits 2709e20, 5f6d09c)
- RunnerStatus pending count was wrong during runner draining — fixed via workSize() + decrement before emit.
- Exact-hash dedup compared against chunk contentHash (includes prepended context); switched to bodyHash via new findByBodyHash.
- Chip "Save" routed through runChat — now routes to runCapture.
- Orchestrator emits one event-log entry per phase (issue #13 acceptance).
- applyPersistedState() runs before any service starts.
- Strategy skips LLM when top score < 0.3.
- Drift report skips rehash when vault mtime matches stored mtime (new VaultService.stat()).
- Cleanup: append drift detection (expectedMtime), shared unique() util, exported TINY_VAULT_THRESHOLD, EmptyRetriever warns once.

## Review fixes (commit abfa828)
- RunnerControls: replaced post-construction monkey-patch with an `onRebuild` constructor hook.
- ingest-writer: heading detection no longer requires a trailing newline; added regression test.
- ingest-strategy: comment clarifying findByBodyHash is a byte match, not semantic.
- runIngest: explicit guard for `confirm` on a dedup_ask preview, with test.

## Test plan
- [x] npx vitest run — 491 passing
- [x] npx tsc --noEmit — clean
- [x] npm run build — clean
- [ ] Smoke test in demo-vault (manual, requires desktop Obsidian): cold-start pill, pause persistence, end-to-end capture, rebuild
- [ ] Coordinate with #111 (retrieval stack) — Services.retriever (currently EmptyRetriever stub) is the merge point

## Notes for review
- Merged origin/dev into the branch (commit 5b2de0f) to pick up #109/#110; one settings collision resolved (consolidated alwaysPreview into alwaysPreviewBeforeSave).
- EmptyRetriever is a deliberate stub — replaced by HybridRetriever once #111 lands.
- Settings tab is sync; meta is pre-loaded into a snapshot via refresh().

### DIFF
--- a/__mocks__/obsidian.ts
+++ b/__mocks__/obsidian.ts
@@ -12,3 +12,22 @@ export class Plugin {}
 export class WorkspaceLeaf {}
 export class ItemView {}
 export class MarkdownRenderer {}
+
+// Minimal stand-ins for tests that reference UI types but don't actually
+// render. Real Obsidian provides full implementations at runtime.
+export class Modal {
+  contentEl: { empty: () => void; createEl: () => unknown } = {
+    empty: () => {},
+    createEl: () => ({}),
+  };
+  open(): void {}
+  close(): void {}
+}
+export class PluginSettingTab {}
+export class Setting {
+  setName(_n: string): this { return this; }
+  setDesc(_d: string): this { return this; }
+  addToggle(_cb: (...args: unknown[]) => unknown): this { return this; }
+  addButton(_cb: (...args: unknown[]) => unknown): this { return this; }
+  addText(_cb: (...args: unknown[]) => unknown): this { return this; }
+}
--- a/prompts/dedup-decider.md
+++ b/prompts/dedup-decider.md
@@ -1,18 +1,28 @@
-version: 0.1.0
+version: 0.2.0
 
 # dedup-decider
 
-Given a candidate note and its top-k similar existing notes, decides
-the dedup strategy: propose new, append to existing, or split into
-multiple notes.
+You compare a candidate note against the user's most-similar existing
+notes and pick a storage strategy.
 
-## Input
-- Candidate NoteSpec.
-- Top-k similar notes with similarity scores.
+## Decisions
+
+- `new` — the candidate is genuinely distinct, even if related notes
+  exist. Default to this when in doubt.
+- `append` — the candidate naturally extends one specific existing note
+  (the user is adding to an ongoing log, project page, or person profile).
+  Provide its `targetPath`.
+- `ask_user` — there is a near-duplicate but you can't tell whether the
+  user wants to merge or branch off. Provide the closest match in
+  `targetPath`. The app will surface a confirmation modal.
+
+## Inputs
+
+- `candidate.title`, `candidate.summary`, `candidate.tags`
+- `similar[]`: ranked existing notes with `path`, `title`, `score`, and a
+  short `snippet`.
 
 ## Output
-JSON with `strategy` (one of `new`, `append`, `split`, `ask_user`),
-`targetPath` (for append), and brief reasoning shown to the user.
 
-## Status
-Scaffold. Exact wording filled in by a separate task.
+Return ONLY a JSON object: `{ "strategy": "new"|"append"|"ask_user",
+"targetPath"?: string, "reason": string }`. No prose, no code fences.
--- a/prompts/ingest-parser.md
+++ b/prompts/ingest-parser.md
@@ -1,16 +1,31 @@
-version: 0.1.0
+version: 0.2.0
 
 # ingest-parser
 
-Turns raw user content into a structured NoteSpec candidate.
+You convert raw user content into a structured note candidate (a NoteSpec).
+Your output is consumed by a deterministic writer; off-schema responses are
+rejected.
 
-## Input
-- Raw user content (Markdown, plain text, or pasted snippet).
-- Optional user instruction (e.g. "save this as a meeting note").
+## Goals
+
+1. Pick a concise, descriptive `title` (≤ 80 chars).
+2. Choose a `type` from: source, evergreen, project, meeting, person,
+   concept. Default to "source" when unsure.
+3. Extract `tags` (lowercase, kebab-case) and `entities` (proper nouns).
+4. Write a 1–3 sentence `summary` and a few `key_points`.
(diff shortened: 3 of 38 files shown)


iamaxeloberg (commented):
Overall solid — the orchestrator design is clean and the test coverage is thorough. A few things to address:

---

**`services/index.ts` — monkey-patching `runnerControls.rebuild` is fragile**

```ts
const origRebuild = runnerControls.rebuild.bind(runnerControls);
runnerControls.rebuild = async () => { ... };
```

Overwriting an instance method after construction breaks if `rebuild` is ever called via the prototype. Better: pass `epochCell` (or an update callback) into `RunnerControls` at construction so it can update the cell itself — or accept a post-rebuild hook in the constructor options.

---

**`ingest-writer.ts` — heading detection has a trailing-newline edge case**

`existing.includes(`\n${heading}\n`)` won't match if the target ends with the heading but no trailing newline. Low risk in practice (Obsidian writes trailing newlines), but a test for that edge case or a looser check (`.includes('\n' + heading)`) would be safer.

---

**`ingest-strategy.ts` — bodyHash exact-match assumes byte-identical LLM output**

`findByBodyHash` will miss a true duplicate if the LLM normalises whitespace differently across two calls. Fine for v1, but worth a comment so a future reader doesn't treat it as a true semantic dedup.

---

**`runIngest` — no test for `confirm` on a `dedup_ask` preview**

The `preview_unresolved` guard is correct, but there's no test for the case where the modal returns `{action: "confirm"}` for a dedup kind. Either a test or an earlier guard in the loop that explicitly rejects `confirm` when `strategy.kind === "dedup_ask"` would close the loop.

---

**Merge order with #111**

`Services.retriever` is stubbed here as `EmptyRetriever`; #111 sets it to `HybridRetriever`. Whoever merges second needs to overwrite. Suggest #111 merges first since it owns the retrieval stack — then this branch just drops the `EmptyRetriever` on rebase.

---

The monkey-patch is the only thing I'd want fixed before merge. Everything else is low-risk or comment-worthy. The PreviewHandler injection, alwaysPreview bypass rules, mtime drift guard, per-phase event log, and AbortSignal threading are all well done.

heimdallSWE:
Thanks for the review — all four code findings addressed in abfa828.

1. **Monkey-patch on `RunnerControls.rebuild`** — replaced with a typed `RunnerControlsOptions.onRebuild` hook passed at construction. `services/index.ts` now wires the epoch-cell refresh through that option. No more post-construction overwriting; prototype dispatch works as expected.
2. **Heading detection trailing-newline edge case** — switched to `\n${heading}` (without the trailing `\n`). Added a regression test for the "target ends at the heading line, no trailing newline" shape.
3. **`findByBodyHash` byte-vs-semantic match** — added a comment explaining the trade-off: byte match is the fast path; LLM normalisation differences fall through to the threshold + similarity branches, which is the intended degradation.
4. **No test for `confirm` on `dedup_ask`** — added explicit guard inside the preview loop (`confirm_on_dedup_ask` failure reason) plus a test that asserts the matched note is unchanged and the queue is untouched.

#5 (merge order with #111) — agreed. Happy to rebase off #111 once it lands; the only collision is the `EmptyRetriever` in `services/index.ts` which gets dropped in favour of `HybridRetriever`.

491 tests pass, typecheck and build clean.

PR #114, opened 2026-05-07
feat(classifier): golden set, eval runner, regression gate (#85 #86 #87)

iamaxeloberg:
## Summary

Scope-bounded revision of the original classifier milestone PR. The runtime classifier service / skip router / threshold gate / disambiguation chip / inspector pieces were superseded by parallel work that landed on dev (#103, #106, #107, #108, #110), so this PR drops them to avoid a contract regression and ships only the eval-side deliverables.

- **#85 — Golden set.** 30 hand-labeled examples in `eval/classifier-golden/examples.json` covering all four labels and every spec edge case (attachment-only, follow-up turns, active-file context, ambiguous phrasing, leading-`?`, recentTurns context).
- **#86 — Eval runner.** `eval/runner/classify.ts` drives dev's real `classifyTurn` orchestrator (skip router → LLM → retries → thresholds → events). Emits confusion matrix, per-class P/R/F1, and P50/P95 latency to `eval/runs/<timestamp>/report.{md,json}`.
- **#87 — Regression gate.** `eval:classifier:ci` runs the scripted-LLM mock against the golden set through the real orchestrator and compares to a committed baseline. P0 fails CI on any rise in ask→capture FPR; P1 warns on >5pp accuracy drop. Workflow at `.github/workflows/classifier-regression.yml`.
- **Committed thresholds** documented in `planning/classifier.md` (capture 0.85 / mixed 0.75 / ask 0.70 / meta 0.70) keyed to dev's `DEFAULT_CLASSIFIER_THRESHOLDS`.

### Baseline

Locked in `eval/classifier-golden/baseline-mock.json`:

| Metric | Value |
|---|---|
| Accuracy | 24 / 30 (80%) |
| Capture precision | 75% |
| Capture recall | 100% |
| ask→capture FPR | 0% |

### What this does NOT include

The previous revision shipped duplicate runtime code under `src/services/{ollama,mock}-classifier.ts` plus a divergent `src/contracts/classifier.ts`. Dev now has the full classifier surface (`classifier-orchestrator`, `classifier-skip-router`, `classifier-thresholds`, `classifier-llm`, `classifier-retries`, `classifier-events`, `classifier-utils`, `classifier-prompt`) plus the disambig chip and dev-mode inspector via separate PRs. Issues #57 and #84 should be closed against those.

## Test plan

- [x] `npm test` — 48 files / 503 tests green
- [x] `npx tsc --noEmit` — clean
[...]

PR #115, opened 2026-05-07
fix(demo): citation chips, classifier silent fallback, 2s timeout

iamaxeloberg:
## Summary

Demo-hardening fixes on top of the classifier milestone (#114). Merge #114 first.

- **Classifier silent fallback** — timeout/parse failures now route to `ask` without showing the disambiguation chip (previously the chip appeared on every slow startup, looking broken).
- **Timeout 500ms → 2000ms** — more realistic for demo hardware where Ollama may be loading the model.
- **`ClassifierDecision.failed` flag** — distinguishes classifier failure from a legitimately low-confidence result; `belowThreshold()` never triggers the chip on failures.
- **Citation "Källor:" strip** — after each assistant response, shows pill badges for every retrieved note. Addresses the citation accuracy metric in the demo evaluation.

## Test plan

- [ ] Start Ollama cold (model not loaded) — send a message — no disambiguation chip appears, answer streams normally
- [ ] Ask about Jonas Berg's hiking trip — answer appears with `Källor: vandringsnoter_sundsvall_harnosand` pill visible
- [ ] Ask about books read — `Källor: laslogg_2025, bokrecension_*` pills shown
- [ ] `npm test` — 120 tests, all green

### DIFF
--- a/src/contracts/classifier.ts
+++ b/src/contracts/classifier.ts
@@ -8,6 +8,8 @@ export interface ClassifierDecision {
   skipReason?: string;
   latencyMs: number;
   promptVersion: string;
+  /** True when the classifier failed (timeout/parse error) and fell back to the default label. */
+  failed?: boolean;
 }
 
 export interface ClassifyOptions {
--- a/src/services/ollama-classifier.ts
+++ b/src/services/ollama-classifier.ts
@@ -8,7 +8,7 @@ import type {
 import { DEFAULT_THRESHOLDS } from "../contracts/classifier";
 
 const PROMPT_VERSION = "v1.0";
-const TIMEOUT_MS = 500;
+const TIMEOUT_MS = 2000;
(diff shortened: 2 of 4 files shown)


SimonProjekt (approved):
Clean fixup PR — the three changes are all correct and well-scoped.

**TIMEOUT_MS 500 → 2000** — fixes the main bug from #114, good.

**Silent fallback with `failed` flag** — logic in `belowThreshold` is right: `if (decision.failed) return false` means failures never show the chip. The optional field works fine here.

**Citation chips** — one minor note: the hover style implies the chips are clickable, but they aren't. Either add `app.workspace.openLinkText(name, "")` on click to navigate to the note, or drop the hover effect to avoid misleading the user. Not blocking for demo purposes.

Approving as a stacked fix on top of #114.

PR #120, opened 2026-05-08
feat(settings): add chat model selector (default gemma4:e4b)

heimdallSWE:
## Summary
- Adds a `chatModel` setting (default `gemma4:e4b`) used by both the chat view and the intent classifier.
- Replaces the `pickDefaultModel` prefix-match heuristic in `view.ts` with an explicit read from settings, so model selection is deterministic and user-controllable.
- Exposes the model as a text field under Settings → Gemmera → Capture.
- Bumps the hard-coded fallbacks in `OllamaLLMService` and `classifier-llm.ts` from `gemma3:latest` to `gemma4:e4b`.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `vitest` — classifier-llm / classifier-retries / classifier-orchestrator suites pass (59 tests)
- [ ] Smoke: rebuild + copy into `demo-vault/.obsidian/plugins/gemmera/`, toggle plugin, confirm "Chat model" field appears and persists; chat uses the configured tag (verify via Verbose logs or Ollama request log)
- [ ] Verify `ollama pull gemma4:e4b` is required and that the chat falls back gracefully if the tag is missing

## Notes
- Chat view caches the model on open; changing the setting requires reopening the Gemmera leaf (or reloading the plugin) to take effect. Happy to make it per-turn if reviewers prefer.
- Classifier model is intentionally tied to the same setting — a single source of truth keeps chat and classifier in sync.

### DIFF
--- a/src/services/classifier-llm.test.ts
+++ b/src/services/classifier-llm.test.ts
@@ -78,7 +78,7 @@ function makeMockLLM(responses: MockResponse[]): LLMService {
       // Verify classifier-specific options are passed.
       expect(opts.format).toBe("json");
       expect(opts.stream).toBe(false);
-      expect(opts.model).toBe("gemma3:latest");
+      expect(opts.model).toBe("gemma4:e4b");
 
       return { content: resp.content ?? "" };
     },
--- a/src/services/classifier-llm.ts
+++ b/src/services/classifier-llm.ts
@@ -14,7 +14,7 @@ import { ChatMessage, LLMService } from "../contracts/llm";
 import { PromptLoader } from "../contracts/prompts";
 import { assembleClassifierPrompt } from "./classifier-prompt";
 
-const CLASSIFIER_MODEL = "gemma3:latest";
+const CLASSIFIER_MODEL = "gemma4:e4b";
 const CLASSIFIER_TIMEOUT_MS = 500;
 
 /** Errors the classifier call surface can throw. */
--- a/src/services/classifier-llm.ts
+++ b/src/services/classifier-llm.ts
@@ -12,9 +12,9 @@
 import { ClassifierInput, ClassifierOutput } from "../contracts/classifier";
 import { ChatMessage, LLMService } from "../contracts/llm";
 import { PromptLoader } from "../contracts/prompts";
+import { DEFAULT_CHAT_MODEL } from "../settings";
 import { assembleClassifierPrompt } from "./classifier-prompt";
 
-const CLASSIFIER_MODEL = "gemma4:e4b";
 const CLASSIFIER_TIMEOUT_MS = 500;
 
 /** Errors the classifier call surface can throw. */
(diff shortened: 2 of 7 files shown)


iamaxeloberg (commented):
## Review

Overall this is a clean PR — the settings plumbing is correct end-to-end and removing the racy `pickDefaultModel()` async probe is a nice simplification. Two things should be fixed before merge:

### Must fix

**1. Potential build error in `src/services/index.ts`**
If `OllamaClassifierService` is still exported from `index.ts` but `ollama-classifier.ts` was deleted on this branch, `tsc` will fail. Please confirm `index.ts` was updated (or that `tsc --noEmit` comes back clean on this branch).

**2. Duplicated default string in `ollama-llm.ts`**
`DEFAULT_MODEL` in `ollama-llm.ts` is set independently to `"gemma4:e4b"` rather than importing from `settings.ts`. These will silently diverge on the next model bump. Suggest:
```ts
// src/services/ollama-llm.ts
import { DEFAULT_CHAT_MODEL } from "../settings";
const DEFAULT_MODEL = DEFAULT_CHAT_MODEL;
```

### Nice to have

- **Silent reset UX** — if the user clears the model input, the value snaps back to `DEFAULT_CHAT_MODEL` but the field visually stays empty. A `text.setValue(DEFAULT_CHAT_MODEL)` after the snap-back would close that gap.
- **Test coverage** — `classifier-llm.test.ts` doesn't exercise the explicit `deps.model = "custom-tag"` override path. One extra test case would cover it.
- **PR description note** — "requires reopening the leaf" isn't accurate; since `this.settings` is a shared object reference the new model is picked up on the next turn without a reopen.

Happy to approve once the two must-fixes are confirmed.

iamaxeloberg:
## PR #120 Review — `feat(settings): add chat model selector`

**Overall:** Core design is sound, settings plumbing works for chat + classifier. All 614 tests pass, compiles cleanly. A few gaps to address before merge:

### 🔴 Critical

**Ingest/capture pipeline misses `chatModel`** — `view.ts:runCapture()` calls `runIngest()` but doesn't pass `model: this.settings.chatModel` in deps. This means every capture/ingest operation silently falls back to `ingest-parser.ts`'s hard-coded `"gemma3:latest"`, ignoring the user's model choice. Inconsistent with chat + classifier behavior.

### 🟠 Major

1. **Duplicate default in `ollama-llm.ts`** — `const DEFAULT_MODEL = "gemma4:e4b"` lives independently from `DEFAULT_CHAT_MODEL` in `settings.ts`. These will silently diverge on the next bump. Import `DEFAULT_CHAT_MODEL` instead.

2. **`query-orchestrator.ts` bypasses the setting** — line 129 still calls `deps.llm.pickDefaultModel()` as fallback. If wired up in the future, it ignores the user's `chatModel` preference.

### 🟡 Minor

1. **Empty-field UX** — clearing the text field snaps back to `DEFAULT_CHAT_MODEL` in settings but the field stays visually empty. Add `text.setValue(DEFAULT_CHAT_MODEL)` after the reset.

2. **`ingest-parser.ts` fallback stale** — hard-coded `"gemma3:latest"` on line 58. Should import from `DEFAULT_CHAT_MODEL`.

3. **Test gap** — `classifier-llm.test.ts` doesn't exercise the custom `deps.model` override path; `makeMockLLM` hard-asserts `"gemma4:e4b"`.

### ✅ What works
- Removes racy `pickDefaultModel()` async probe on view open
- Good single-source-of-truth pattern with `DEFAULT_CHAT_MODEL`
- Clean JSDoc on `model?` parameters

### Priority fix order
1. Thread `chatModel` into `runIngest()` deps
2. Import `DEFAULT_CHAT_MODEL` in `ollama-llm.ts`
3. Address `query-orchestrator` stale call (or document gap)
4. Fix empty-field UX + test coverage

PR #122, opened 2026-05-08
feat(tool-loop): mixed-intent orchestration — ingest then query in one turn (#47)

### LINKED ISSUE #47: Tool-loop: mixed-intent orchestration (ingest then query in one turn)
Handle the `mixed` classifier label by chaining the ingest and query state machines within a single user turn.

### Acceptance
- On `mixed`, ingest path runs first to completion; query path runs second with the newly-written note included in the retrieval set.
- A single `turn_id` spans both phases; events are tagged with a `phase` field (`ingest` | `query`).
- The UI shows both phases as distinct status cards within one turn.
- Cancellation during ingest aborts the whole turn; cancellation after ingest preserves the saved note and aborts only the query phase.
- Hard stops (#4) apply across the combined turn, not per phase.
- Test covers: clean mixed turn, mixed cancelled mid-ingest, mixed cancelled mid-query.

### Reference
- `planning/tool-loop.md` — "Mixed intent".
- `planning/classifier.md` — `mixed` taxonomy entry.

SimonProjekt:
## Summary

- Closes #47
- Adds `runMixed` orchestrator that chains `runIngest` → `runQuery` under a single `turnId`
- `EventLogEntry` gains an optional `phase: "ingest" | "query"` field for turn inspector phase separation
- `QueryOrchestratorDeps` gains `onStateChange` so the query phase drives live status updates
- `view.ts` routes `intent === "mixed"` to a new `runMixed` method with two phase-specific status elements
- Cancellation during ingest aborts the whole turn; aborting `querySignal` after ingest preserves the saved note

## Test plan

- [ ] 6 new tests in `src/services/mixed-orchestrator.test.ts` — all pass (`620/620` suite green)
- [ ] Clean mixed turn saves note and returns answer with citations
- [ ] Ingest cancel propagates as `{ kind: "cancelled", phase: "ingest" }`, query never starts
- [ ] Pre-aborted `querySignal` yields `{ kind: "failed", phase: "query" }`, job queue confirms note was saved
- [ ] `onStateChange` called with correct phase for both ingest and query states
- [ ] TypeScript compiles clean (`tsc --noEmit`)

### DIFF
--- a/src/contracts/state-machine.ts
+++ b/src/contracts/state-machine.ts
@@ -132,6 +132,8 @@ export interface EventLogEntry {
   fromState: string | null;
   timestamp: number;
   triggeringEvent: StateMachineEvent | null;
+  /** Set on mixed-intent turns to distinguish which phase produced the entry. */
+  phase?: "ingest" | "query";
 }
 
 // Implementations must guarantee that any entry passed to a fully-resolved
--- a/src/services/mixed-orchestrator.test.ts
+++ b/src/services/mixed-orchestrator.test.ts
@@ -171,44 +171,64 @@ describe("runMixed", () => {
     expect(queryEntries).toHaveLength(0);
   });
 
-  it("cancelled mid-query via signal: note is saved, query phase fails gracefully", async () => {
-    const queryReply = JSON.stringify({ answer: "Answer.", citations: [] });
-
+  it("cancelled mid-query via signal (pre-abort): note is saved, query fails with savedPath", async () => {
     const queryController = new AbortController();
-    const vault = new MockVaultService({});
-    const store = new InMemoryIngestionStore();
-    const queue = new InMemoryJobQueue();
-    const writer = new IngestWriter(vault);
-    const eventLog = new InMemoryEventLog();
-
-    // Abort the query signal immediately
     queryController.abort();
 
-    const deps: MixedOrchestratorDeps & { eventLog: InMemoryEventLog } = {
-      llm: new SequenceLLM([specReply(), queryReply]),
-      promptLoader,
-      retriever: new StubRetriever([makeHit("note.md")]),
-      store,
-      vault,
-      writer,
-      jobQueue: queue,
-      preview: confirmPreview,
-      assembler: new StubAssembler([makeChunk("note.md")]),
-      eventLog,
-      turnId: "turn-2",
-      alwaysPreview: false,
-      querySignal: queryController.signal,
-    };
+    const deps = setup({ llmReplies: [specReply()] });
+    deps.querySignal = queryController.signal;
 
     const outcome = await runMixed("test", deps);
 
-    // Query phase failed (aborted) but ingest note was preserved
     expect(outcome.kind).toBe("failed");
     if (outcome.kind !== "failed") return;
     expect(outcome.phase).toBe("query");
+    expect(outcome.savedPath).toBeTruthy();
 
-    // Verify the job queue has an index job for the saved note (ingest completed)
-    const jobs = queue.drain();
+    const jobs = (deps.jobQueue as InMemoryJobQueue).drain();
+    expect(jobs.some((j) => j.kind === "index")).toBe(true);
+  });
+
+  it("in-flight query abort: note is saved, query phase fails after LLM aborted", async () => {
+    const queryController = new AbortController();
+
+    // First LLM call (ingest parse): resolves immediately.
+    // Second call (query generate): hangs until signal is aborted.
+    let calls = 0;
+    const blockingLlm: LLMService = {
+      async chat(opts: ChatOptions): Promise<LLMResponse> {
+        if (calls++ === 0) return { content: specReply() };
+        return new Promise((_, reject) => {
+          if (opts.signal?.aborted) {
+            reject(new DOMException("Aborted", "AbortError"));
+            return;
+          }
+          opts.signal?.addEventListener("abort", () =>
+            reject(new DOMException("Aborted", "AbortError")),
+          );
+        });
+      },
+      async isReachable(): Promise<LLMReachability> { return "running"; },
+      async listModels(): Promise<string[]> { return []; },
+      async pickDefaultModel(): Promise<string> { return "test-model"; },
+    };
+
+    const deps = setup({ llmReplies: [] });
+    deps.llm = blockingLlm;
+    deps.querySignal = queryController.signal;
+
+    // Abort after ingest completes but while query is in flight
+    setTimeout(() => queryController.abort(), 0);
+
+    const outcome = await runMixed("test", deps);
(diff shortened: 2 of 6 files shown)


iamaxeloberg (commented):
## Review

The orchestration design is solid — the two-signal cancellation model (ingest-aborts-whole-turn vs. query-aborts-preserves-note) is correctly implemented and well-tested, and the `PhaseEventLog` decorator is a clean seam. Four things should be fixed before merge:

### Must fix

**1. `model` not forwarded in `view.ts:runMixed`**
`runChat` passes `model: this.model` but `runMixed` doesn't. Mixed turns fall back to two extra `pickDefaultModel()` round-trips and always use the default model — a behavioural regression vs. the other turn types. Add `model: this.model` to the `MixedOrchestratorDeps` object in `runMixed`.

**2. Unreachable `else ""` branch — `mixed-orchestrator.ts:100–105`**
By this point `"cancelled"` and `"failed"` have already returned early, so the final `else ""` for `savedPath` can never be reached. Either remove it or add a `satisfies never` assertion so future additions to `IngestOutcome` are caught at compile time.

**3. `split_saved` silently drops all paths after the first — `mixed-orchestrator.ts:104`**
`ingestOutcome.paths[0] ?? ""` is surfaced to the user as if only one note was created. The capture path handles `split_saved` with a full list. Either surface all paths here too, or add a comment explaining why only the first is used.

**4. `StubAssembler` signature too narrow — `mixed-orchestrator.test.ts:47`**
The stub ignores `hits` entirely (the primary input to a real assembler). The stub should declare all three parameters (using `_hits` prefix) so it documents the contract and won't silently drift if `runQuery` starts relying on `hits`.

### Nice to have

- **No test for `validation_failed` query outcome** — the branch exists in `view.ts:287–289` but has no coverage. Worth one extra test.
- **`onStateChange` non-null assertions** — capturing `const { onStateChange } = deps` before the ternary would avoid the `!` assertions on the adapter lambdas.
- **CSS specificity** — `.gemmera-mixed-status--ingest/--query` wins `padding-left` only by source order. Merging it into the base rule would make it less fragile.
- **`recentTurns` updated on cancellation** — if intentional (matches `runCapture` pattern), a brief comment would prevent future confusion.
- **`mixed-orchestrator` not exported from `index.ts`** — inconsistent with other services; worth aligning or documenting the omission.

Happy to approve once the four must-fixes are in.

iamaxeloberg:
## PR #122 Review — `feat(tool-loop): mixed-intent orchestration`

**Author:** SimonProjekt | **Files:** 6 (+493 / -0) | **Base:** `dev`

### Summary

Chains `runIngest` → `runQuery` under a single `turnId`. Clean design: an optional `phase` field on `EventLogEntry` tags both phases in the event log via a `PhaseEventLog` wrapper. The `view.ts` renders two live status elements (one per phase). Cancellation during ingest aborts the whole turn; abort after ingest preserves the note and only fails the query. 6 new tests pass (620/620 green). TypeScript compiles clean.

---

### 🔴 Critical

**1. Saved note is silently lost when query phase throws unexpected error**

`view.ts:runMixed()` wraps the entire `runMixed()` call in try/catch. If `runMixed` throws during the query phase (after ingest succeeded and the note was saved to disk), the catch block shows `appendErrorMessage(err, ...)` with no mention that the note was actually saved. The user might re-try the same message, creating a duplicate note.

**Fix:** Track `savedPath` in a scoped variable outside the try, or store it on `this` so the catch block can inform the user: "Note saved to X, but query failed: ..."

---

### 🟠 Major

**2. `split_saved` with empty `paths` yields empty-string `savedPath`**

In `runMixed()` (mixed-orchestrator.ts:109-113):
```ts
const savedPath = ingestOutcome.kind === "split_saved"
  ? (ingestOutcome.paths[0] ?? "")
  : ...;
```

If `paths` is unexpectedly empty, `savedPath` becomes `""`. In `view.ts`, this produces `"Gemmera: saved to "` notice and `"Saved to ****"` message. While unlikely in practice, the fallback should be defensive.

**Fix:** Fall back to something like `"(split into multiple notes)"` instead of empty string.

---

**3. No test for in-flight query signal abort**

The test "cancelled mid-query via signal" pre-aborts the `AbortController` before calling `runMixed`. This exercises the `deps.querySignal?.aborted` fast-path but not the case where abort fires *during* `runQuery` execution. Since `runQuery` accepts `signal`, it should handle this, but without a test, it's unverified.

---

**4. Query phase structured failures don't mention that note was already saved**

In `view.ts:runMixed()`, `outcome.kind === "failed"` with `phase === "query"` shows `"Answer failed: ${outcome.reason}"`. But the note was already saved in the ingest phase. The user sees only an error and has no idea their note is intact.

**Fix:** Append "Note was saved to X" in the failure message text.

---

### 🟡 Minor

**5. Unnecessary non-null assertions**

Two instances of `deps.onStateChange!(...)` in `mixed-orchestrator.ts` (lines 90 and 144). The ternary guard on `deps.onStateChange` already narrows the type — the `!` is redundant.

---

**6. Missing test for `version` passthrough to `runIngest`**

`MixedOrchestratorDeps.version` is forwarded to `runIngest` but no test validates this wiring. Low risk, but a regression could silently drop the version tag used for observability.

---

**7. Duplicate `turn-2` hard-coded test turnId**

The "cancelled mid-query via signal" test uses `turnId: "turn-2"` inline instead of using the shared `setup()` helper. This avoids a `turnId` collision with "turn-1" (used in other tests), but breaks the pattern and adds a silent constraint on `setup()` test IDs.

---

### ✅ What's Working Well

- ✅ Clean separation: `runMixed` orchestrates, `PhaseEventLog` tags, `view.ts` renders.
- ✅ 6 tests cover happy path, ingest cancel, query signal abort, phase tagging, and ingest failure.
- ✅ Backward-compatible: `phase?` on `EventLogEntry` is optional.
- ✅ `onStateChange` callback pattern matches existing `runIngest` convention.
- ✅ `PhaseEventLog` wrapper is a nice design — zero-cost abstraction.
- ✅ TypeScript compiles cleanly (`tsc --noEmit`), suite passes 620/620.

---

### Recommended Priority

| # | Severity | What |
|---|----------|------|
| 1 | 🔴 | Track savedPath in catch block for unexpected query-phase errors |
| 2 | 🟠 | Show saved note path on structured query-phase failures |
| 3 | 🟠 | Defensive fallback for empty `path[0]` on split |
| 4 | 🟠 | Add test for in-flight abort of query signal |
| 5 | 🟡 | Drop unnecessary `!` assertions |

---

**Verdict:** Solid architecture — the chain, phase tagging, and cancellation semantics are well thought-out. The main gap is UX when query fails after a successful save: the user needs to know their note was saved. Fix items #1 and #2, and this is ready to merge.

PR #135, opened 2026-05-14
feat(rag): query tool loop — ask intent through hybrid retriever + synthesis notes (#14)

### LINKED ISSUE #14: Tools: query tool loop (search, rerank, answer, optional synthesis)

heimdallSWE:
## Summary

Closes #14. Wires the `ask` intent through the RAG query tool loop and adds the optional synthesis-note write-back.

Before this PR, `ask`-tagged turns went through `runChat`, which used the week-2 `VaultLinearIndexService` (a linear text scan). The hybrid retriever + reranker + payload assembler from #8/#9/#10 were only invoked by `runMixed`. After this PR, `ask` turns go through `runQuery` like `mixed` does, and the linear scanner is scoped to `meta`/fallback only.

### Changes

- **`runAsk` in `src/view.ts`** drives `runQuery` with the same services `runMixed` uses (`retriever`, `payloadAssembler`, `eventLog`). Dispatch in `handleSend` and the disambiguation chip "Ask" action now both route through it.
- **`src/services/synthesis-writer.ts`** — `createSynthesisNote` builds a `NoteSpec` with `cowork.source: "synthesis"`, citations in `related[]`, and a `## Sources` wikilink section, then writes via the existing `IngestWriter` so the frontmatter contract is shared.
- **"Save answer as note" button** rendered under answered turns; on click writes a synthesis note using `turnId` as `run_id`.
- **`VaultLinearIndexService` docstring** narrowed to its remaining `meta`/fallback role.
- **Retrieval invariant test** — `SpyRetriever` asserts `retriever.retrieve` is called with the user query for the happy path, closing the "log the state without doing the work" loophole the event-log-only assertion left open.

### Acceptance criteria (issue #14)

- [x] Vault questions never bypass retrieval — invariant test in `query-orchestrator.test.ts` + comment pinning the dispatch in `view.ts`.
- [x] Answers include source paths in the UI payload — citation chips render on the answered branch of `runAsk`.
- [x] Synthesis notes carry the same `cowork` block + frontmatter contract as ingest output — `createSynthesisNote` reuses `IngestWriter`; test asserts `cowork_managed: true` and `source: synthesis` in the written file.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] `npx vitest run` — 635/635 pass (+7 new: 6 synthesis-writer, 1 retrieval invariant)
- [ ] Smoke in demo vault: ingest a note, ask a question that targets it, verify citation chips and retrieval status chip render
- [ ] Smoke synthesis: click "Save answer as note", verify the produced file in `Inbox/` has `cowork.source: synthesis` and a `## Sources` section

## Notes / follow-ups

- `VaultLinearIndexService` kept in `createServices` because `runChat` still uses it for `meta` turns. Removal would change behavior for non-vault paths; out of scope here.
- Synthesis notes land in `settings.inboxFolder` (default `Inbox/`). A dedicated `Synthesis/` folder setting was considered but deferred to avoid expanding the settings surface.
- Reranker not wired into `runAsk` because `services.reranker` is not yet exposed; query-orchestrator handles its absence gracefully. Follow-up issue welcome.

### DIFF
--- a/src/services/index.ts
+++ b/src/services/index.ts
@@ -224,6 +224,7 @@ export { RunnerControls } from "./runner-controls";
 export { ScheduledReconciler } from "./scheduled-reconciler";
 export { IngestWriter } from "./ingest-writer";
 export { runIngest } from "./ingest-orchestrator";
+export { createSynthesisNote, buildSynthesisSpec } from "./synthesis-writer";
 export type {
   IngestPreview,
   PreviewDecision,
--- a/src/services/query-orchestrator.test.ts
+++ b/src/services/query-orchestrator.test.ts
@@ -22,6 +22,15 @@ class StubRetriever implements Retriever {
   }
 }
 
+class SpyRetriever implements Retriever {
+  readonly calls: Array<{ query: string; topK: number | undefined }> = [];
+  constructor(private readonly hits: RetrievalHit[] = []) {}
+  async retrieve(query: string, opts?: { topK?: number }): Promise<RetrievalHit[]> {
+    this.calls.push({ query, topK: opts?.topK });
+    return this.hits;
+  }
+}
+
 class SequenceLLM implements LLMService {
   private calls = 0;
   constructor(private readonly replies: string[]) {}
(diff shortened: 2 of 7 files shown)


SimonProjekt (approved):
Solid implementation. Clean separation between `buildSynthesisSpec` (pure) and `createSynthesisNote` (I/O) makes the tests easy to read and maintain. The INVARIANT comment in `view.ts` is good documentation — it explains *why* `runAsk` must never fall through to `runChat`.

The `SpyRetriever` test genuinely closes a loophole that the event-log assertion alone left open.

Two minor observations (neither blocks merge):

- `version: input.version ?? "0.0.1"` in `synthesis-writer.ts` is a hardcoded fallback. Not a problem now, but worth tracking as a follow-up so it doesn't silently drift from the actual plugin version.
- The `validation_failed` branch in `runAsk` uses the simpler `appendMessage` path while the answered branch builds DOM nodes manually for citation chips. The inconsistency is probably intentional but isn't obvious at first glance — a short comment would help future readers.

635/635 tests passing. ✓

PR #136, opened 2026-05-14
feat(scope): .coworkignore parser wired into indexer and file-watcher (#29, #30)

### LINKED ISSUE #29: Scope: gitignore-style .coworkignore parser with hard-coded exclusions
Parse `vault/.coworkignore` as gitignore-style globs. Supported syntax: `folder/`, `*.glob`, `**` deep wildcard, leading `!` negation, `# comment` lines, blank lines.

Evaluation is **last-match-wins**, consistent with gitignore. Hard-coded exclusions (`.obsidian/`, `.git/`, `.trash/`, `.coworkmd/`) are always applied first and cannot be un-excluded by `!` rules.

Expose a single `isIgnored(vaultRelativePath: string): boolean`.

### Acceptance
- `Templates/` excludes the folder and everything under it.
- `*.canvas` excludes canvas files at any depth.
- `Templates/**` plus `!Templates/KeepMe.md` keeps the negated file.
- `# comment` lines and blank lines are ignored.
- Hard-coded paths return true even if the user writes `!.obsidian/`.
- Default contents (Templates/, Attachments/, attachments/, assets/, *.canvas, *.excalidraw) ship at first install.

### Reference
planning/runtime.md → ".coworkignore" → "Format" and "Default contents on first install".

### LINKED ISSUE #30: Scope: wire .coworkignore into indexer and file-watcher pipeline
Wire `isIgnored()` into the indexer and file-watcher pipeline so matched files are skipped end-to-end:

- **Indexing**: matched files never get chunked, embedded, or inserted into `notes`/`chunks`.
- **Retrieval**: because they never enter the index, they never appear in results (no extra filter needed).
- **File watcher**: `vault.on('create' | 'modify' | 'delete' | 'rename')` events for matched paths are dropped before reaching the `jobs` table.

Hard-coded exclusions are checked here too as a defense-in-depth.

### Acceptance
- Adding a path to `.coworkignore` then editing a matching file triggers no jobs.
- Cold-start scan respects the ignore file from the first walk.
- Renaming an ignored file into a non-ignored path triggers a normal create-equivalent index job.
- Renaming a non-ignored indexed file into an ignored path removes its rows.

### Reference
planning/runtime.md → ".coworkignore" → "What it affects" and "What it does not affect".

SimonProjekt:
## Summary

- **`CoworkIgnore` parser** (`src/services/cowork-ignore.ts`): gitignore-style syntax with last-match-wins evaluation. Supports folder patterns (`Templates/`), globs (`*.canvas`, `**`), negation (`!`), comments (`#`), and blank lines. Hard-coded system paths (`.obsidian/`, `.git/`, `.trash/`, `.coworkmd/`) are always ignored and cannot be negated.
- **Default `.coworkignore`**: created on first install with `Templates/`, `Attachments/`, `attachments/`, `assets/`, `*.canvas`, `*.excalidraw`.
- **Wired into `DefaultPathFilter`**: indexer, reconciler, and vault-event-bridge all respect the rules from cold start (`createServices` reads the file before starting any services).
- **Live reload**: editing `.coworkignore` in Obsidian reloads the rules immediately via `registerEvent` — no plugin restart needed.

## Tests (32 new)

Covers: parsing (blank lines, comments), folder patterns, glob patterns (`*`, `**`, `?`), negation with last-match-wins, hard ignores (all four system paths), `reload()`, default contents, anchored vs unanchored patterns.

## Test plan

- [ ] Create `Templates/Weekly.md` and verify it is never sent to the job queue
- [ ] Add a custom pattern to `.coworkignore`, edit a matching file, verify no index job fires (live reload)
- [ ] Rename a note from an ignored path to a non-ignored path → index job fires
- [ ] Rename a note from a non-ignored path to an ignored path → delete job fires
- [ ] Fresh vault: verify `.coworkignore` is created with default contents on first activation

Closes #29, closes #30.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -42,6 +42,18 @@ export default class GemmeraPlugin extends Plugin {
     // process jobs the user expected to stay paused.
     await this.services.runnerControls.applyPersistedState();
     this.services.eventBridge.start();
+    // Reload .coworkignore rules whenever the file is saved, so the user does
+    // not need to restart Obsidian after editing it.
+    this.registerEvent(
+      this.app.vault.on("modify", async (file) => {
+        if (file.path === ".coworkignore") {
+          try {
+            const content = await this.app.vault.adapter.read(".coworkignore");
+            this.services.coworkIgnore.reload(content);
+          } catch { /* file removed; keep current rules */ }
+        }
+      }),
+    );
     this.services.ingestionRunner.start();
     this.services.embeddingService.start();
     this.services.bm25IndexService.start();
--- a/src/services/cowork-ignore.test.ts
+++ b/src/services/cowork-ignore.test.ts
@@ -1,0 +1,175 @@
+import { describe, expect, it } from "vitest";
+import { CoworkIgnore, DEFAULT_COWORKIGNORE } from "./cowork-ignore";
+
+// ── Parsing ───────────────────────────────────────────────────────────────────
+
+describe("CoworkIgnore — parsing", () => {
+  it("ignores blank lines", () => {
+    const ig = new CoworkIgnore("\n\n   \n");
+    expect(ig.matches("Notes/foo.md")).toBe(false);
+  });
+
+  it("ignores comment lines", () => {
+    const ig = new CoworkIgnore("# this is a comment\n  # also a comment");
+    expect(ig.matches("Notes/foo.md")).toBe(false);
+  });
+});
+
+// ── Folder patterns ───────────────────────────────────────────────────────────
+
+describe("CoworkIgnore — folder patterns (trailing /)", () => {
+  const ig = new CoworkIgnore("Templates/");
+
+  it("ignores the folder itself", () => {
+    expect(ig.matches("Templates")).toBe(true);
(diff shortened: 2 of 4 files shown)


SimonProjekt:
32 tests for a parser is thorough coverage. Hard-ignoring system paths (`.obsidian/`, `.git/` etc.) with no negation override is the right call.

Live reload via `registerEvent` is correctly wired — the listener is cleaned up automatically on plugin unload. The nested try-catch for the `.coworkignore` write fallback (read-only vault) is defensive without being over-engineered.

One behavioral difference from true gitignore worth noting: directory patterns like `Templates/` are effectively anchored to the vault root in this implementation (`path.startsWith(rule.pattern + "/")`). Real gitignore would also match `nested/Templates/foo.md`. For a typical Obsidian flat-folder structure this doesn't matter, but it's worth a comment in the default `.coworkignore` or a follow-up issue so users aren't surprised if they have e.g. `Projects/Templates/` and expect the pattern to catch it.

Issue #53, opened 2026-04-27
NotePreviewModal + tool dispatcher (save_note create)


heimdallSWE:
M1 / PR 4 — `ui/03-preview-modal`. Build the preview gate so no LLM-driven vault write happens silently.

Includes:
- `NotePreviewModal` (Obsidian `Modal`) with editable title, folder path (defaults to `Inbox/`), frontmatter form, Markdown body preview via `MarkdownRenderer.renderMarkdown`.
- `tool-dispatcher.ts`: when the LLM emits a `tool_call` for `save_note` with `mode: "create"`, the dispatcher opens the modal, awaits Confirm/Cancel, then invokes `VaultService.createFile` on confirm.
- Cancel drops the tool call and appends a system message ("Save cancelled.") to the chat.
- `tool-dispatcher.ts` is the **only** caller of `VaultService` mutation methods (enforced by `no-restricted-imports` lint rule banning write methods elsewhere — risk mitigation from §Risks).
- Manual test checklist appended to `README.dev.md`.

### Acceptance
- "create a note about dogs" → mock emits `save_note` tool call → preview modal appears → Confirm writes file under `Inbox/`.
- Cancel drops the tool call and posts the system message; no file is written.
- Lint fails if any file outside `tool-dispatcher.ts` imports a `VaultService` write method.

### Reference
`planning/ui-plugin-plan.md` §M1 (PR 4) and §Risks "Preview gate bypass".

iamaxeloberg:
**Deferred to v2.** Pairs with #52 (NotePreviewModal). v0.1 capture goes through #55 (direct-save with Undo) — no modal, no full dispatcher. Reopen alongside #52 when preview flow is scoped.

iamaxeloberg:
**Reopened — closed in error.** PR #116 (merged to dev) implements the preview modal + tool dispatcher (save_note + ingest path) via `src/ui/ingest-preview-modal.ts` and the ingest orchestrator (`src/services/ingest-orchestrator.ts`, `ingest-writer.ts`). My v2 deferral was wrong.

PR #137, opened 2026-05-14
feat(dev): turn inspector modal behind devMode setting (#74)

### LINKED ISSUE #74: Dev-mode turn inspector modal
M3 / PR 13 — `ui/11-turn-inspector`. Dev-mode introspection so we can debug a turn end-to-end.

Includes:
- Dev-mode-only command `cowork.open-turn-inspector`.
- Modal showing, per turn: user input, classifier verdict, tool calls in order with args + results, model raw output, final rendered answer, elapsed time per step.
- Readable JSON dump with copy-to-clipboard.

### Acceptance
- Command is hidden unless dev-mode is enabled in settings.
- After any turn, opening the inspector shows the full trace including tool args/results and per-step timing.
- Copy-to-clipboard yields valid JSON.

### Reference
`planning/ui-plugin-plan.md` §M3 (PR 13).

SimonProjekt:
## Summary

- **`devMode` setting**: new boolean in `GemmeraSettings` (default `false`), toggled via Settings → Advanced → "Developer mode". Persists in `data.json`.
- **Command**: `open-turn-inspector` registered with `checkCallback` — hidden from the command palette unless `devMode` is `true`. No command appears for production users.
- **Enhanced `TurnInspectorModal`**:
  - Elapsed time column: milliseconds since the previous state entry
  - "Copy JSON" button: writes the full raw `EventLogEntry[]` (untruncated payloads) to the clipboard as formatted JSON; shows "Copied!" / "Copy failed" feedback for 1.5 s
- **`GemmeraChatView`**: tracks `lastTurnId` at the start of every turn so the inspector always shows the most recent trace; `openTurnInspector()` fetches entries from the in-memory event log and opens the modal

## Test plan

- [ ] Enable Developer mode in Settings → Advanced
- [ ] Send a chat or capture message
- [ ] Open command palette → "Turn Inspector — view last turn trace" → modal appears with state table, elapsed times, and Copy JSON button
- [ ] Click "Copy JSON" → paste into a JSON validator → valid JSON with full payloads
- [ ] Disable Developer mode → command disappears from palette
- [ ] Open inspector before any turn → notice "No turn recorded yet"

Closes #74.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -113,6 +113,24 @@ export default class GemmeraPlugin extends Plugin {
     });
 
     this.addCommand({
+      id: "open-turn-inspector",
+      name: "Turn Inspector — view last turn trace",
+      checkCallback: (checking: boolean) => {
+        if (!this.settings.devMode) return false;
+        if (!checking) {
+          const leaves = this.app.workspace.getLeavesOfType(VIEW_TYPE);
+          const view = leaves[0]?.view as GemmeraChatView | undefined;
+          if (view) {
+            view.openTurnInspector();
+          } else {
+            new Notice("Gemmera: Open the chat panel first.");
+          }
+        }
+        return true;
+      },
+    });
+
+    this.addCommand({
       id: "capture-selection",
       name: "Gemmera: Fånga markering",
       hotkeys: [{ modifiers: ["Ctrl", "Shift"], key: "C" }],
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -48,6 +48,13 @@ export interface GemmeraSettings {
 
   /** Ollama tag used for chat and the classifier (e.g. `gemma4:e4b`). */
   chatModel: string;
+
+  /**
+   * When true, the "Open turn inspector" command appears in the palette.
+   * Shows raw state traces with tool args, payloads, and per-step timing.
+   * Intended for developers; hidden by default in production.
+   */
+  devMode: boolean;
 }
 
 export const DEFAULT_CHAT_MODEL = "gemma4:e4b";
@@ -62,4 +69,5 @@ export const DEFAULT_SETTINGS: GemmeraSettings = {
   ollamaPath: "",
   turnTimeoutMs: 120_000,
   chatModel: DEFAULT_CHAT_MODEL,
+  devMode: false,
 };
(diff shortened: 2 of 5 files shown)


heimdallSWE (commented):
Clean, well-scoped change. Hides the inspector behind a setting, adds elapsed-time column, adds Copy JSON. The `checkCallback` pattern is the right way to gate command visibility. Mostly nits — one concrete UX bug.

## Bug

**1. The "Reload the plugin after toggling" hint in the setting description is wrong.**
```ts
checkCallback: (checking) => {
  if (!this.settings.devMode) return false;
  ...
}
```
`checkCallback` is invoked every time the palette opens, so it re-reads `this.settings.devMode` live. Toggling the setting reflects in the palette on the next open with no reload needed. Either drop the hint, or — if there's a real reason it doesn't update live in your testing — investigate the actual cause (cached settings reference?) before papering over with a reload instruction.

## Nits

**2. `lastTurnId` is process-local.**
After a plugin reload, the inspector for the previously-completed turn returns no entries (the `InMemoryEventLog` is gone). The "send a message first" notice covers the empty case correctly, but worth a sentence in the modal when entries are empty post-reload, e.g. "No entries — the event log was cleared on the last plugin reload."

**3. Inline styles deepen an existing inconsistency.**
The original `turn-inspector.ts` already used inline styles; this PR adds more (`contentEl.style.maxWidth = "860px"`, `marginBottom`, `whiteSpace`, gap, etc.). The rest of the codebase puts styling in `styles.css` (see the `.gemmera-note-preview__*` block from #53). Not a regression, but a good follow-up issue: move the inspector to a `.gemmera-turn-inspector__*` class set.

**4. Hard pixel cap at 860px.**
`contentEl.style.maxWidth = "860px"` ignores small viewports. `max-width: min(860px, 95vw)` (in CSS) is friendlier on narrow Obsidian windows.

**5. `setTimeout` after `navigator.clipboard.writeText` can fire after modal close.**
Harmless (writing `textContent` on a detached node is a no-op), but for hygiene you can store the timer and clear it in `onClose`. Skip if you'd rather not add bookkeeping for a 1.5s window.

LGTM modulo the misleading reload hint.

PR #138, opened 2026-05-14
fix(scope): gitignore-correct .coworkignore semantics + working live reload

heimdallSWE:
## Summary

Follow-up to #136 fixing two issues found in post-merge review:

### 1. Live reload was dead code
`app.vault.on("modify", …)` never fires for root-level dotfiles — Obsidian hides them from the `Vault` API. The handler in `main.ts` never executed.

**Fix:** parent-dir `fs.watch` filtered for `.coworkignore` (parent-dir watch survives atomic-save inode swaps that single-file watches miss), debounced 200ms, plus a `Gemmera: Reload .coworkignore` command for manual triggering.

### 2. Directory patterns deviated from gitignore semantics
`Templates/` was always anchored to the vault root, so `Sub/Templates/x.md` was silently *not* ignored under the default config. Per gitignore, a pattern with no internal slash is unanchored and matches at any depth.

**Fix:** unify `parseRule`/`testRule` through a single regex builder with a `(?:/.*)?` tail for dir patterns. As a side benefit:
- `/Templates/` correctly anchors to root (was previously also unanchored — bug)
- Glob dir patterns like `Draft-*/` now work
- Removes the special-case dir-matching branch

## Behavior after fix

| Pattern | `Templates/x.md` | `Sub/Templates/x.md` | `Templates` |
|---|---|---|---|
| `Templates/` | ✓ | ✓ (new) | ✓ |
| `/Templates/` | ✓ | ✗ (now correct) | ✓ |
| `Archive/2024/` | n/a | ✗ | n/a |
| `Draft-*/` | n/a | ✓ (new) | n/a |

## Tests
5 new (37 total, all green):
- unanchored dir matches at any depth
- leading-slash anchoring
- internal-slash anchoring
- glob dir patterns
- CRLF line endings

Full suite: 685/685 passing, `tsc` clean.

## Test plan

- [ ] Create `Sub/Templates/note.md` under default config → verify it is **not** queued for indexing
- [ ] Edit `.coworkignore` in an external editor → rules reload within ~200ms (check console)
- [ ] Run `Gemmera: Reload .coworkignore` from command palette → rules reload immediately
- [ ] Delete `.coworkignore` → no crash; existing rules retained

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,5 +1,5 @@
 import { FileSystemAdapter, Menu, Notice, Plugin, TFile, WorkspaceLeaf } from "obsidian";
-import { readFileSync, existsSync } from "fs";
+import { readFileSync, existsSync, watch as fsWatch } from "fs";
 import { GemmeraChatView, VIEW_TYPE } from "./view";
 import { createServices, Services } from "./services";
 import { OllamaLifecycle, type OllamaStatus } from "./services/ollama-lifecycle";
@@ -42,18 +42,7 @@ export default class GemmeraPlugin extends Plugin {
     // process jobs the user expected to stay paused.
     await this.services.runnerControls.applyPersistedState();
     this.services.eventBridge.start();
-    // Reload .coworkignore rules whenever the file is saved, so the user does
-    // not need to restart Obsidian after editing it.
-    this.registerEvent(
-      this.app.vault.on("modify", async (file) => {
-        if (file.path === ".coworkignore") {
-          try {
-            const content = await this.app.vault.adapter.read(".coworkignore");
-            this.services.coworkIgnore.reload(content);
-          } catch { /* file removed; keep current rules */ }
-        }
-      }),
-    );
+    this.registerCoworkIgnoreReload();
     this.services.ingestionRunner.start();
     this.services.embeddingService.start();
     this.services.bm25IndexService.start();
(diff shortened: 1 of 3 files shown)


SimonProjekt (approved):
Direkt svar på de två anmärkningarna från granskningen av #136. Båda rätt fixade.

**Live reload (main.ts)**

`fs.watch` på vault-rooten är rätt lösning — parent-dir watch överlever atomiska sparanden (inode-swap). Cleanup hanteras korrekt: både debounce-timern och watchern stängs via `this.register()` vid plugin-avlastning. `persistent: false` är med.

One thing to be aware of: on some Linux filesystems `fs.watch` may return `null` as the `filename` argument, causing the handler to silently skip the reload (`null !== ".coworkignore"` → `return`). In practice this is fine for atomic saves (which trigger `rename` events that always carry the filename), and the command palette fallback covers the rest. Not a blocker.

**Regex refactor (cowork-ignore.ts)**

Logic is correct and simpler than the original — the separate `testRule` + dir-matching branch are gone, replaced by a single regex with a `(?:/.*)?` tail for directory patterns. Verified all four cases from the PR table:

- `Templates/` → `(?:^|/)Templates(?:/.*)?$` — matches at any depth ✓
- `\/Templates/` → `^Templates(?:/.*)?$` — anchored to root ✓
- `Archive/2024/` → `^Archive/2024(?:/.*)?$` — anchored via internal slash ✓
- `Draft-*/` → `(?:^|/)Draft-[^/]*(?:/.*)?$` — glob + any depth ✓

CRLF tolerance is implicit via `raw.trim()` in `parseRule` — the test confirms it.

685/685 green.

PR #139, opened 2026-05-14
feat(ui): tool dispatcher + NotePreviewModal for save_note create (#53)

### LINKED ISSUE #53: NotePreviewModal + tool dispatcher (save_note create)

SimonProjekt:
## Summary
- **New** `src/services/tool-dispatcher.ts`: routes `LLMToolCall` objects to the right handler; #53 implements `save_note(mode=create)`
- **New** `src/ui/note-preview-modal.ts`: Obsidian Modal with editable title, folder path (defaults to `Inbox/`), comma-separated tags, and `MarkdownRenderer` body preview; resolves a Promise so the dispatcher can await the user's decision
- **Modified** `src/view.ts`: passes `SAVE_NOTE_TOOL` to `llm.chat()` in `runChat` and dispatches any returned `toolCalls` through `dispatchToolCall`; cancel appends a "Save cancelled." system message
- **Modified** `styles.css`: `.gemmera-note-preview__*` rules for the new modal

## Test plan
- [ ] With `llmBackend: "mock"` send a message containing "create" — mock emits `save_note` tool call → preview modal appears
- [ ] Fill in title / folder / tags and click **Save** → note written under `Inbox/`
- [ ] Click **Cancel** → no file written; chat shows "Save cancelled."

Closes #53

### DIFF
--- a/src/services/tool-dispatcher.test.ts
+++ b/src/services/tool-dispatcher.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, it, vi } from "vitest";
+import { MockVaultService } from "../contracts/mocks/mock-vault";
+import { dispatchToolCall, buildFrontmatter, type ToolDispatchDeps } from "./tool-dispatcher";
+
+function makeDeps(overrides: Partial<ToolDispatchDeps> = {}): ToolDispatchDeps {
+  const vault = new MockVaultService();
+  return {
+    vault,
+    inboxFolder: "Inbox/",
+    openNotePreview: vi.fn().mockResolvedValue(null),
+    appendSystemMessage: vi.fn(),
+    ...overrides,
+  };
+}
+
+describe("dispatchToolCall — routing", () => {
+  it("routes save_note with mode=create to save handler", async () => {
+    const confirmResult = { confirmed: true as const, title: "Dogs", folder: "Inbox/", tags: [] };
+    const deps = makeDeps({ openNotePreview: vi.fn().mockResolvedValue(confirmResult) });
+
+    const result = await dispatchToolCall(
+      { id: "1", name: "save_note", arguments: { mode: "create", title: "Dogs", body_markdown: "Dogs are great." } },
+      deps,
+    );
+
+    expect(result.kind).toBe("done");
+    expect(await deps.vault.exists("Inbox/Dogs.md")).toBe(true);
+  });
+
+  it("returns unknown_tool for save_note with mode=append (not yet handled)", async () => {
+    const deps = makeDeps();
+    const result = await dispatchToolCall(
+      { id: "2", name: "save_note", arguments: { mode: "append" } },
+      deps,
+    );
+    expect(result.kind).toBe("unknown_tool");
+  });
+
+  it("returns unknown_tool for unrecognised tool names", async () => {
+    const deps = makeDeps();
+    const result = await dispatchToolCall(
+      { id: "3", name: "hallucinated_tool", arguments: {} },
+      deps,
+    );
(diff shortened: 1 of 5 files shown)


heimdallSWE (commented):
Foundational PR — the dispatcher surface looks right. A few things to address before merge, plus context that applies to the whole stack (#141, #142).

## Stack ordering

This PR and #141/#142 are a flat stack (each contains the previous as a commit), but all target `dev` standalone. They must merge in order. After this squash-merges, #141 and #142 will need rebases — flagging so it doesn't catch us off guard.

## Bugs

**1. `ingestWriter` is declared in `ToolDispatchDeps` but never used.**
`dispatchSaveNoteCreate` writes directly via `deps.vault.create`. Either route through `IngestWriter` (consistent with how the rest of the codebase writes + ingests in one step) or drop the dep until #141 actually needs it.

**2. No collision handling — `vault.create` throws on existing path.**
If the LLM proposes "Meeting Notes" and that file exists, `vault.create` throws after the user already confirmed the modal. With no try/catch in `view.ts` around `dispatchToolCall`, the error breaks the entire message handler. Either check `vault.exists` and suffix with `(2)`, or catch and surface via `appendSystemMessage`.

**3. Frontmatter is not YAML-safe.**
`` `title: "${title}"` `` breaks on titles containing `"`, `\`, or starting with YAML indicators (`>`, `|`, `&`, `*`). Use `JSON.stringify(title)` (valid YAML for strings) at minimum.

## Robustness — applies to all three PRs in the stack

- **No try/catch around `dispatchToolCall` in `view.ts`.** Any vault throw kills the message handler. Wrap with a `try` that logs + `appendSystemMessage("Tool call failed: …")`.
- **`unknown_tool` silently dropped** in the view loop. Surface an "unknown tool" notice so hallucinated tool names don't fail invisibly.
- **Zero tests.** Dispatcher routing and `buildFrontmatter` are both pure-ish and easy to unit test — would unblock confidence for the rest of the stack.

## Nits

- Modal: doubled assignment (`value:` in `createEl` *and* `titleInput.value = ...` after) — pick one.
- Modal: no Enter-to-save keybinding; standard Obsidian pattern.
- Modal: empty title silently falls back to `this.opts.title` — consider disabling Save instead.

Approving the direction; would like the YAML escaping and collision handling resolved before merge.

PR #141, opened 2026-05-14
feat(ui): tool dispatcher — all write tools (#62)

### LINKED ISSUE #62: Tool dispatcher: all write tools + non-overridable DeleteConfirmModal
M2 / PR 8 — `ui/06-tool-dispatcher-all-writes`. Extend the dispatcher to all write tools and ship the non-overridable delete gate.

Includes:
- `tool-dispatcher.ts` handles: `save_note(mode=append)`, `update_frontmatter`, `rename_or_move_note`, `delete_note`, `create_synthesis_note`.
- `DeleteConfirmModal` as a separate modal class — required even when "Always preview" is off. Includes note title, path, and a typed "delete" confirmation for friction.
- Append mode follows the dated-heading convention from `tool-loop.md`.
- Rename/move uses `FileManager.renameFile` so incoming links update atomically.

### Acceptance
- Each of the six write tools fires through the correct preview path (preview modal or delete modal).
- Delete forces typed confirmation even with "Always preview" disabled.
- Rename updates incoming `[[wikilinks]]` automatically (verified against a test note with backlinks).
- Append targets the correct dated heading or creates it.

### Reference
`planning/ui-plugin-plan.md` §M2 (PR 8). Visuals/inline-preview surface live in UI-surfaces v1 / PR 7 wide-mode (separate issue).

SimonProjekt:
## Summary
Extends `tool-dispatcher.ts` from #53 to handle all six write tool calls that the LLM can emit:

| Tool | Handler | Modal |
|---|---|---|
| `save_note(mode=create)` | `dispatchSaveCreate` | `NotePreviewModal` (editable) |
| `save_note(mode=append)` | `dispatchSaveAppend` | none (direct write under dated heading) |
| `update_frontmatter` | `dispatchUpdateFrontmatter` | none (read-patch-write) |
| `rename_or_move_note` | `dispatchRename` → `runDestructiveOp` | `RenamePreviewModal` |
| `delete_note` | `dispatchDelete` → `runDestructiveOp` | `DeleteConfirmModal` (non-overridable) |
| `create_synthesis_note` | `dispatchSynthesis` → `createSynthesisNote` | none |

- Cancel on any write tool posts "X cancelled." as a system message
- `view.ts` wires `openDeleteConfirm` and `openRenamePreview` as UI callbacks in `dispatchDeps`
- All write tools exposed as `WRITE_TOOLS` array (replaces single `SAVE_NOTE_TOOL`)

## Test plan
- [ ] Mock LLM: send "rename notes/a.md to notes/b.md" → rename modal appears, link count shown, confirm renames file
- [ ] Mock LLM: send "delete notes/x.md" → delete modal appears regardless of Always-preview setting
- [ ] Cancel on delete → no file removed; chat shows "Delete cancelled."

> **Note:** This branch is based on `feature/tool-dispatcher-53` (#139). Merge #139 first.

Closes #62

### DIFF
--- a/src/contracts/mocks/mock-vault.ts
+++ b/src/contracts/mocks/mock-vault.ts
@@ -32,6 +32,11 @@ export class MockVaultService implements VaultService {
     this.files.set(path, content);
   }
 
+  async modify(path: string, content: string): Promise<void> {
+    if (!this.files.has(path)) throw new Error(`File not found: ${path}`);
+    this.files.set(path, content);
+  }
+
   /** Test-only convenience: overwrite an existing file or create a new one. */
   setFile(path: string, content: string): void {
     this.files.set(path, content);
--- a/src/contracts/vault.ts
+++ b/src/contracts/vault.ts
@@ -19,6 +19,7 @@ export interface VaultService {
   read(path: string): Promise<string>;
   exists(path: string): Promise<boolean>;
   create(path: string, content: string): Promise<void>;
+  modify(path: string, content: string): Promise<void>;
   append(path: string, content: string): Promise<void>;
   getHeadings(path: string): Promise<HeadingRef[]>;
   /**
--- a/src/services/links-index-service.ts
+++ b/src/services/links-index-service.ts
@@ -61,6 +61,10 @@ export class LinksIndexService {
     return this.inFlight;
   }
 
+  neighborCount(path: string): number {
+    return this.index.neighborCount(path);
+  }
+
   private schedule(event: RunnerEvent): void {
     const next = this.inFlight.then(() => this.process(event));
     this.inFlight = next.catch(() => undefined);
--- a/src/services/real-vault.ts
+++ b/src/services/real-vault.ts
@@ -26,6 +26,11 @@ export class ObsidianVaultService implements VaultService {
     await this.app.vault.create(path, content);
   }
 
+  async modify(path: string, content: string): Promise<void> {
+    const file = this.requireFile(path);
+    await this.app.vault.modify(file, content);
+  }
+
   async append(path: string, content: string): Promise<void> {
     const file = this.requireFile(path);
     await this.app.vault.append(file, content);
--- a/src/services/tool-dispatcher.test.ts
+++ b/src/services/tool-dispatcher.test.ts
@@ -1,13 +1,22 @@
 import { describe, expect, it, vi } from "vitest";
 import { MockVaultService } from "../contracts/mocks/mock-vault";
-import { dispatchToolCall, buildFrontmatter, type ToolDispatchDeps } from "./tool-dispatcher";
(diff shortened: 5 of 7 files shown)


heimdallSWE (commented):
This adds five write tools on top of #139. The shape is right (delegating to `runDestructiveOp` for destructive ops is the right call), but several handlers have real bugs that will bite the first time the LLM exercises them.

## Critical

**1. `patchFrontmatter` regex injection / breakage.**
```ts
const re = new RegExp(`^${key}:.*$`, "m");
```
`key` is LLM-supplied. If the LLM passes a key like `foo.bar`, `foo[1]`, or anything with regex metacharacters, the regex matches the wrong thing or throws. Escape it:
```ts
const escaped = key.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```

**2. `dispatchUpdateFrontmatter` writes via `vault.create` to overwrite.**
The inline comment says "VaultService.create overwrites if file exists" — that contradicts `contracts/vault.ts` and Obsidian's own `app.vault.create` (which throws on existing files). Either verify the implementation actually overwrites, or switch to `vault.modify`. If the assumption is wrong, this tool throws on every call.

**3. `patchFrontmatter` silently no-ops on unclosed frontmatter.**
If `raw.indexOf("---", 3) === -1`, returns `raw` unchanged with no error. Dispatcher then reports "Updated frontmatter in X" — false positive. Detect this case and return `cancelled` with an explanation.

**4. `appendUnderDatedSection` always prepends a new `## YYYY-MM-DD` heading.**
Five appends in a day → five identical H2 sections in the file. Scan the file for today's heading and append underneath it.

**5. Rename preview shows `linkCount: 0` always.**
```ts
{ kind: "rename", from: args.from, to: args.to, affectedLinkCount: 0 }
```
This is the number shown to the user in `RenamePreviewModal`. The codebase already has `services.linksIndexService`; pass real count or the preview misleads the user.

## Smaller

**6. `dispatchSynthesis` uses `crypto.randomUUID()` as `runId`.**
That's not a runId — it ties the synthesis note to nothing. Either pass an actual run identifier from the chat turn or use a more honest placeholder (`` `tool-call-${Date.now()}` ``).

**7. `JSON.stringify(value)` is not YAML.**
Mostly OK for scalars and string arrays, but nested objects produce JSON-shape output rather than YAML block form — readable but inconsistent. For null/undefined/dates the result is silently wrong (date-as-quoted-string). Document the limitation or use a YAML serializer.

**8. `vault.append` doesn't normalize trailing newline.**
If the existing file ends without `\n`, you get `last-line\n\n## DATE`. Minor but easy to fix.

## Tests

Still nothing under `*.test.ts`. `patchFrontmatter` is pure and would benefit most — every bullet above could be a one-liner test.

PR #143, opened 2026-05-14
feat(ui): chat history drawer — session list, auto-title, retention (#70)

### LINKED ISSUE #70: Chat history drawer + DuckDB-shaped JSON persistence + retention
M2 / PR 11 — `ui/09-chat-history-drawer`. Persist chats and expose them via a history drawer with auto-titling and retention policy.

Includes:
- `HistoryDrawer.svelte` lists prior chats with title, timestamp, turn count.
- Persistence v1: JSON file at `vault/.coworkmd/chats.json`. The persistence API surface is **DuckDB-shaped** so the M3 swap to `chats.duckdb` is internal.
- Chat retention setting: N days / N chats / unlimited (default unlimited).
- Auto-title from first user message.

### Acceptance
- Closing and reopening Obsidian restores the chat list.
- Selecting a prior chat rehydrates the `ChatStore` correctly.
- Retention pruning runs on plugin load and respects the user's choice.
- Auto-title appears immediately after the first user message is sent.

### Reference
`planning/ui-plugin-plan.md` §M2 (PR 11) and §Dependencies (`chats.duckdb` fallback).

SimonProjekt:
## Summary
Wires the existing `ChatHistoryStore` (already in place for persistence since the previous sprint) into a visible history drawer in the right-side context panel.

- **Session list**: sorted by last-updated; shows auto-generated title (first 50 chars of user's first message), relative date, and message count
- **Switch session**: clicking a session replays its turns into the message area and rehydrates `this.history` so the next send continues in context
- **New chat**: "+ Ny chatt" button clears the current conversation and resets session state
- **Live updates**: drawer re-renders after each turn so title/count update immediately
- **Retention**: `pruneIfNeeded()` runs on panel open (currently unlimited — configurable via `retentionPolicy()`)
- **Layout**: drawer sits inside the existing `.gemmera-context-panel` which is hidden at < 600px and appears on the right at ≥ 600px (no layout change needed)
- **Storage**: JSON at `vault/.coworkmd/chats.json` (DuckDB-shaped API surface per spec — the store interface doesn't change when the backend is swapped)

## Test plan
- [ ] Send 2–3 messages → session appears in drawer with title derived from first message
- [ ] Reload Obsidian → chat list restored from `chats.json`
- [ ] Click a prior session → messages replay; next send continues that session
- [ ] "+ Ny chatt" → blank slate; prior sessions still visible in drawer
- [ ] Panel hidden at narrow viewport, visible at ≥ 600px

Closes #70

### DIFF
--- a/src/services/chat-history.ts
+++ b/src/services/chat-history.ts
@@ -67,6 +67,7 @@ export class ChatHistoryStore {
   }
 
   async pruneIfNeeded(): Promise<void> {
+    if (this.retention.maxDays === undefined && this.retention.maxSessions === undefined) return;
     const data = await this.load();
     let sessions = [...data.sessions];
     if (this.retention.maxDays !== undefined) {
--- a/src/view.ts
+++ b/src/view.ts
@@ -45,6 +45,7 @@ export class GemmeraChatView extends ItemView {
   private escCleared = false;
   private prefersReducedMotion = false;
   private lastTurnId: string | undefined;
+  private historyDrawerEl: HTMLElement | null = null;
 
   constructor(
     leaf: WorkspaceLeaf,
@@ -139,12 +140,18 @@ export class GemmeraChatView extends ItemView {
     });
 
     this.contextPanelEl = bodyEl.createEl("div", { cls: "gemmera-context-panel" });
+    this.historyDrawerEl = this.contextPanelEl;
 
     const adapter = this.app.vault.adapter as FileSystemAdapter;
     const historyPath = adapter.getFullPath(".coworkmd/chats.json");
-    this.chatHistory = new ChatHistoryStore(historyPath);
+    this.chatHistory = new ChatHistoryStore(historyPath, this.retentionPolicy());
     // Sessions are created lazily on the first persisted turn so opening the
     // panel without sending anything doesn't accumulate empty rows in chats.json.
+
+    // Prune stale sessions on open, then render the drawer.
+    this.chatHistory.pruneIfNeeded().catch(() => {}).finally(() => {
+      this.renderHistoryDrawer().catch(() => {});
+    });
   }
 
   async onClose(): Promise<void> {
@@ -529,6 +536,7 @@ export class GemmeraChatView extends ItemView {
           const sid = this.currentSessionId;
           await ch.appendTurn(sid, { role: "user", content: text, timestamp: userTs });
           await ch.appendTurn(sid, { role: "assistant", content: reply.content, timestamp: Date.now() });
+          await this.renderHistoryDrawer();
         })().catch(() => {});
       }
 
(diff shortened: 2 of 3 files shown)


heimdallSWE (commented):
Drawer wiring looks right and the lazy-session pattern composes cleanly with the existing `appendTurn` flow. A couple of concrete bugs and a few correctness/architecture notes.

## Bugs

**1. `historyDrawerEl = this.contextPanelEl` aliases the whole context panel.**
```ts
this.contextPanelEl = bodyEl.createEl("div", { cls: "gemmera-context-panel" });
this.historyDrawerEl = this.contextPanelEl;
```
`renderHistoryDrawer` then calls `el.empty()` on every render. Today nothing else lives in the context panel, so this is safe — but it silently claims the whole pane. Any future feature that adds content to the panel will be wiped on the next render. Suggest:
```ts
this.historyDrawerEl = this.contextPanelEl.createEl("div", { cls: "gemmera-history" });
```

**2. `switchToSession` doesn't reset `recentTurns`.**
`this.history` (LLM context) is rebuilt from `session.turns`, but `recentTurns` (classifier context, view.ts:492) keeps the previous session's last 3 entries. Switching from "help me refactor this code" → a journaling session leaves stale `intent` signals for the classifier, biasing routing on the first turn of the new session. Reset `this.recentTurns = []` in `switchToSession`.

**3. `pruneIfNeeded()` writes the file even when retention is empty.**
`ChatHistoryStore.pruneIfNeeded` always calls `await this.flush({ sessions })` whether or not anything was pruned. With `retentionPolicy()` returning `{}`, this is an unconditional read+write of `chats.json` on every panel open. Short-circuit:
```ts
async pruneIfNeeded(): Promise<void> {
  if (this.retention.maxDays === undefined && this.retention.maxSessions === undefined) return;
  ...
}
```

**4. `formatSessionMeta` divides turn count by 2.**
```ts
const turns = Math.floor(turnCount / 2);
```
This assumes paired user/assistant turns. If a user message ever lacks a response (error, cancel, in-flight) the count is off by one or rounds down silently. Count user turns directly: `session.turns.filter((t) => t.role === "user").length`.

## Architecture / UX

**5. `renderHistoryDrawer()` rebuilds the entire list DOM on every turn.**
Fine at small N; at hundreds of sessions you'll see flicker on every send. Consider rendering once on open + mutating active state / inserting new sessions diff-style. Not blocking — flag as a follow-up.

**6. No delete-session affordance.**
Sessions accumulate forever (the default policy is empty). Users can't remove an embarrassing chat without editing `chats.json` by hand. Either ship a per-row trash icon now or open a follow-up issue — but document the gap somewhere.

**7. Broad `.catch(() => {})` swallows everything.**
```ts
this.chatHistory.pruneIfNeeded().catch(() => {}).finally(...)
this.renderHistoryDrawer().catch(() => {});
this.switchToSession(session).catch(() => {});
```
A corrupted `chats.json` shows up as "Inga sparade chattar." with no diagnostic. At minimum `.catch((err) => console.error("[gemmera] history drawer:", err))`.

**8. Locale hardcoded to `"sv-SE"` and UI strings are Swedish ("Ny chatt", "Inga sparade chattar.", "meddelande{n}").**
Consistent with `"Öppna chatt"` elsewhere in the codebase, so probably intentional — flagging in case the team wants a single decision on i18n strategy before more Swedish strings ship.

**9. PR description says "DuckDB-shaped API surface per spec" but the store is JSON-backed.**
`ChatHistoryStore` reads/writes `chats.json`. Either the description is aspirational (and should say "JSON for now, DuckDB later") or the implementation drifted from the spec. Worth aligning so reviewers/future-you don't mis-read the contract.

## Smaller

- **`pruneIfNeeded` will silently delete old conversations** when retention is later enabled. Worth a UX note (toast on prune, or "X archived" indicator) before flipping that on.
- **No keyboard nav between sessions** beyond Tab. Up/Down arrow within the list would polish the experience but isn't blocking.
- **Drawer hidden at <600px** (existing media query, not introduced here) — users on narrow windows can't switch sessions. Acceptable, just document.

PR #144, opened 2026-05-14
chore(ui): audit and close responsive-layout (#40)

heimdallSWE:
Audit-only PR for **#40 — Responsive narrow/wide layout via CSS container query**.

All four acceptance criteria in the issue are already met by current \`dev\`. No behavioural change needed; the diff is a three-line CSS comment that anchors the breakpoint choice (container query vs. viewport media query) back to #40 so future maintainers don't reach for \`matchMedia\` by reflex.

## Audit findings

| Acceptance criterion | Where it's satisfied |
|---|---|
| Container query, not media query | \`container-type: inline-size; container-name: gemmera\` at \`styles.css:6-7\`; \`@container gemmera (min-width: 600px)\` at line ~33 |
| Both modes render shared header + composer | \`gemmera-status\` (header) + \`gemmera-input-area\` (composer) mount inside \`gemmera-main\` at \`view.ts:79-104\` — unconditional |
| Wide mode reserves right column | \`.gemmera-context-panel { display: none }\` by default; \`display: flex; flex: 2\` at ≥600 px; main is \`flex: 3\` → 60/40 split per spec |
| No JS resize listener | \`grep -rn "resize\\|ResizeObserver\\|innerWidth\\|clientWidth" src/\` returns nothing. Only \`matchMedia\` use is \`prefers-reduced-motion\` (a11y) |
| Resize crosses breakpoint without scroll jumps | \`gemmera-body\` is \`overflow: hidden; min-height: 0\`; only \`.gemmera-messages\` and \`.gemmera-context-panel\` scroll. Column collapse doesn't remount nodes |

## Out of scope

The issue mentions a "show sources" expander in narrow mode and the drag-drop target swap (composer row in narrow, full view in wide). Neither is in the formal acceptance bullets, so they're not part of this audit. Drag-drop can land alongside the ingest flow PRs (#52 / #55) if we want it before the milestone closes.

## Test plan

- [x] \`npm test\` — 705 passed
- [x] \`npm run build\` — clean
- [ ] Visual check: drag the Gemmera pane across the 600 px width threshold in Obsidian; layout swaps single-column ↔ two-column without scroll jumps
- [ ] Pop the leaf to a separate window; resize the window across 600 px; layout responds independently of the original main window

Closes #40.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #40, opened 2026-04-27
Responsive narrow/wide layout via CSS container query


heimdallSWE:
Implement single-breakpoint responsive layout at container width 600 px using a CSS container query (not viewport media query) so layout responds to pane resize.

**Narrow (<600 px):** single column, messages stacked, citations as compact inline chips, retrieval detail behind a "show sources" expander, ingestion preview as modal, drag-drop target = composer row.

**Wide (≥600 px):** two columns, chat left ~60%, context panel right ~40%. Code blocks/tables/Mermaid render at full width. Drag-drop target = entire view.

### Acceptance
- Resizing the pane crosses the 600 px breakpoint and swaps layout without scroll jumps.
- Both modes render the shared header and composer.
- Wide mode reserves the right column for the context panel (content wired in a later issue).
- No JS resize listener needed — purely CSS container query.

### Reference
`planning/ui-surfaces.md` — "Responsive layout".

iamaxeloberg:
**Deferred to v2 (architectural blocker).** Assumes a Svelte-based responsive layout. Current codebase is vanilla TypeScript/DOM with no CSS container queries on the chat view. Reopen if/when a Svelte migration (#50) is scoped.

iamaxeloberg:
**Reopened — closed in error.** PR #119 (merged to dev) implements responsive layout in `styles.css` with CSS container queries (`container-type: inline-size; container-name: gemmera; @container gemmera (min-width: 600px)`). My "Svelte blocker" framing was wrong — dev did it in plain CSS without any Svelte dependency.

PR #140, opened 2026-05-14
feat(eval): retrieval golden-set harness + regression gate (#16)

heimdallSWE:
Closes #16 — the last open issue in **RAG v1**.

## Summary

- Adds `eval/runner/retrieval.ts` — in-process retrieval-eval harness mirroring the classifier runner. Builds the full RAG retrieval stack against fixture mini-vaults under `evals/golden/fixtures/`.
- Authors 27 new golden examples across three new shards (`lexical`, `semantic`, `mixed`), bringing the set to **30 total** as called out in the issue.
- Tracks Recall@10, Recall@30, MRR, citation correctness, per-`winningSignal` attribution, retrieval P50/P95, and end-to-end P50/P95 latency.
- Persists one row per run to `eval/runs/metrics.jsonl` (intended consumer: the in-app trend chart in UI-surfaces v1).
- Regression gate (`--compare baseline.json`): fails the run on a Recall@10 drop ≥10pp (issue's P0). Wired into CI as `npm run eval:retrieval:ci`.
- Mock-mode baseline: Recall@10 93%, Recall@30 100%, MRR 0.45, Citation 100%.

## Acceptance check (vs. issue)

| Issue | Status |
|---|---|
| 30 examples in golden set | ✅ (sharded — `link-structure` + `lexical` + `semantic` + `mixed`) |
| Recall@10, Recall@30, MRR | ✅ |
| Citation correctness | ✅ (payload self-consistency check) |
| Latency P50/P95 retrieval + end-to-end | ✅ |
| Rule-based faithfulness | ⏸ Deferred — needs the LLM-driven citation output from the query tool loop. Documented inline + in `eval/README.md`. |
| Single command runs the full set | ✅ `npm run eval:retrieval:mock` |
| Deliberate-regression trips the flag | ✅ `eval/runner/regression-gate.test.ts` — pure-function regression-gate test (7 cases) |
| Results persisted for trend lines | ✅ `eval/runs/metrics.jsonl`, append-only |
| 10% recall drop flagged | ✅ FP-tolerant comparison; exits 1 on P0 |

## Out of scope (explicit deferrals)

- LLM-judge faithfulness — `planning/rag.md §Evaluation` defers to v2.
- In-app trend chart — UI-surfaces v1.
- Scheduled weekly run — Runtime v1 candidate; CI step covers the regression-gate case in the meantime.
- Live-mode (real Ollama embedder) runner script — the flag plumbing is there; the npm script is mock-only for now since real-embedder runs aren't safe for CI.

[...]

PR #145, opened 2026-05-14
feat(chat): token streaming + Stop button with immediate cancellation (#46)

### LINKED ISSUE #46: Token streaming + Stop button with immediate cancellation
Stream answers token-by-token via Ollama's SSE endpoint. The Send button becomes a Stop button during generation.

Cancellation behavior: model call aborted immediately; any in-flight tool calls complete but their results are dropped; the partial answer is preserved in history with a "cancelled" marker so the user can edit and retry.

### Acceptance
- Tokens render as they arrive with no perceptible buffering.
- Stop interrupts within ~100 ms of click.
- Cancelled turn is visually distinct and labelled.
- Edit-and-retry from a cancelled turn produces a new turn (does not overwrite).

### Reference
`planning/ui-surfaces.md` — "Streaming and cancellation".

heimdallSWE:
Closes #46. PR 2 of the UI-surfaces v1 sequence (PR 1 = #144 / audit-only #40).

## Summary

- Adds `StreamingState` service (owns the `AbortController` for the in-flight turn) plus 9 unit tests
- Send button doubles as Stop during a turn (`Skicka` ↔ `Stoppa`); single click handler dispatches by state
- `AbortSignal` threaded into `runChat`, `runAsk`, and `runMixed` (orchestrators were already plumbed)
- Cancelled turn renders a "stopped" marker with muted styling and preserves any partial tokens already streamed
- In-flight text captured on send → Stop refills the composer for edit-and-retry → retry produces a *new* turnId, never overwrites

## Acceptance (issue #46)

| Criterion | Status |
|---|---|
| Tokens render as they arrive with no perceptible buffering | ✅ Existing `ollama-llm.ts` streaming preserved; runChat's `onToken` appends to textEl |
| Stop interrupts within ~100 ms of click | ✅ AbortController.abort() is synchronous; fetch read-loop sees the signal on the next chunk boundary |
| Cancelled turn is visually distinct and labelled | ✅ `gemmera-message--cancelled` class (muted + italic + left-border), inline `[stopped]` tag |
| Edit-and-retry produces a new turn | ✅ Each Send creates a fresh `turnId`; the cancelled turn stays in the DOM/history alongside the retry |

## Out of scope

In-flight tool calls intentionally complete on their existing semantics — mixed-orchestrator drops query results when `querySignal` fires; ingest already-written notes stay (reversible via existing Undo). The issue's "in-flight tool calls complete but their results are dropped" matches this behaviour.

## Test plan

- [x] `npm test` → 715/715 (up from 705)
- [x] `npm run build` → clean
- [ ] Manual: send a long-ish query, click Stop, verify the partial answer stays with the stopped tag and Send is restored
- [ ] Manual: stop a turn → composer refills → edit one word → send → new turn appears below the cancelled one
- [ ] Manual: stop during \`runAsk\` (\"Searching notes…\" status) → status replaced by stopped marker; no error toast

### DIFF
--- a/src/services/streaming-state.test.ts
+++ b/src/services/streaming-state.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it } from "vitest";
+import { isAbortError, StreamingState } from "./streaming-state";
+
+describe("StreamingState", () => {
+  it("starts idle", () => {
+    const s = new StreamingState();
+    expect(s.isStreaming()).toBe(false);
+  });
+
+  it("begin() returns an AbortSignal and marks streaming", () => {
+    const s = new StreamingState();
+    const signal = s.begin();
+    expect(signal).toBeInstanceOf(AbortSignal);
+    expect(signal.aborted).toBe(false);
+    expect(s.isStreaming()).toBe(true);
+  });
+
+  it("cancel() aborts the signal and returns true", () => {
+    const s = new StreamingState();
+    const signal = s.begin();
+    expect(s.cancel()).toBe(true);
+    expect(signal.aborted).toBe(true);
+    expect(s.isStreaming()).toBe(false);
+  });
+
+  it("cancel() on idle returns false and is a no-op", () => {
+    const s = new StreamingState();
+    expect(s.cancel()).toBe(false);
+  });
+
+  it("end() clears streaming without aborting the signal", () => {
+    const s = new StreamingState();
+    const signal = s.begin();
+    s.end();
+    expect(s.isStreaming()).toBe(false);
+    expect(signal.aborted).toBe(false);
+  });
+
+  it("begin() while already streaming aborts the previous controller", () => {
+    const s = new StreamingState();
+    const first = s.begin();
+    const second = s.begin();
+    expect(first.aborted).toBe(true);
+    expect(second.aborted).toBe(false);
+    expect(s.isStreaming()).toBe(true);
+  });
+
+  it("subsequent cancel() after begin/cancel is a no-op", () => {
+    const s = new StreamingState();
+    s.begin();
+    s.cancel();
+    expect(s.cancel()).toBe(false);
+  });
+});
+
+describe("isAbortError", () => {
+  it("recognizes DOMException AbortError", () => {
+    const err = new DOMException("Aborted", "AbortError");
+    expect(isAbortError(err)).toBe(true);
+  });
+
+  it("recognizes an Error with name=AbortError", () => {
+    const err = new Error("aborted");
+    err.name = "AbortError";
+    expect(isAbortError(err)).toBe(true);
+  });
+
+  it("rejects non-AbortError errors", () => {
+    expect(isAbortError(new Error("boom"))).toBe(false);
+    expect(isAbortError(new TypeError("nope"))).toBe(false);
+    expect(isAbortError("string")).toBe(false);
+    expect(isAbortError(undefined)).toBe(false);
+  });
+});
--- a/src/services/streaming-state.ts
+++ b/src/services/streaming-state.ts
@@ -1,0 +1,47 @@
+/**
+ * Tracks whether the chat view is currently streaming a response. Owns the
+ * `AbortController` so the Stop button can cancel the in-flight LLM call (and
+ * any orchestrator that threads the same `AbortSignal`).
+ *
+ * Lives as a tiny module so the view's send/stop button toggling can be
+ * exercised in isolation — view-level UI is otherwise untested.
+ */
+export class StreamingState {
+  private controller: AbortController | null = null;
+
+  isStreaming(): boolean {
+    return this.controller !== null;
+  }
+
+  /** Begin a new streaming turn and return its `AbortSignal`. */
+  begin(): AbortSignal {
+    if (this.controller) {
+      // Defensive: caller didn't end the previous turn. Abort it so the user
+      // never has two concurrent in-flight calls racing into the same view.
+      this.controller.abort();
+    }
+    this.controller = new AbortController();
+    return this.controller.signal;
+  }
+
+  /** Cancel the in-flight turn. Returns true if anything was cancelled. */
+  cancel(): boolean {
+    if (!this.controller) return false;
+    this.controller.abort();
+    this.controller = null;
+    return true;
+  }
+
+  /** Mark the turn as finished cleanly (no abort). */
+  end(): void {
+    this.controller = null;
+  }
+}
+
+/**
+ * Pattern shared with `classifier-llm.ts` and `retry-policy.ts` — DOMException
+ * with name "AbortError" plus stdlib errors that carry the same name.
+ */
+export function isAbortError(err: unknown): boolean {
+  return err instanceof Error && err.name === "AbortError";
+}
--- a/src/view.ts
+++ b/src/view.ts
@@ -237,7 +237,7 @@ export class GemmeraChatView extends ItemView {
 
     // ── Mixed (#47) ───────────────────────────────────────────────────
     if (intent === "mixed") {
-      const signal = this.setStreaming(true)!;
+      const signal = this.setStreaming(true);
       try {
         await this.runMixed(text, turnId, signal);
       } finally {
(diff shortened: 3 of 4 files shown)


heimdallSWE (commented):
Clean, well-scoped change. Pulling the abort plumbing into a tiny testable `StreamingState` module is the right move — the unit tests cover the state machine surface cleanly. One concrete bug that contradicts the PR's own comment, plus a behavioral gap worth deciding on before merge.

## Bug

**1. The cancelled-runChat branch pops the user message from `this.history`, contradicting the inline comment.**

```ts
} catch (err) {
  if (isAbortError(err)) {
    // Preserve any partial tokens already rendered into textEl and mark
    // the turn as cancelled. Pop the assistant entry from history because
    // we never received a complete reply; the user text stays so the next
    // turn keeps the right context.
    this.markMessageCancelled(assistantEl, textEl);
    this.history.pop();
  }
```

In the abort path, no assistant entry was ever pushed — `this.history.push({ role: "assistant", ... })` at view.ts:528 only runs after `services.llm.chat` *resolves*. So `pop()` here removes the **user** message that was pushed at line 499. The comment says "the user text stays so the next turn keeps the right context," but the code does the opposite.

User-visible effect: send "What's the capital of France?" → click Stop → send "And Germany?" → the LLM has no record of the cancelled question and treats "And Germany?" as a standalone turn.

Fix: drop the `this.history.pop()` from the cancelled branch. (Keep it in the error branch — that path's behavior matches the comment.)

## Behavioral gap

**2. Stop is unreachable during the classify phase.**

`handleSend` calls `setInputDisabled(true)` (which disables the send button) *before* awaiting `classifyAndDecide`. `setStreaming(true)` — which re-enables the button as Stop — only runs after classification resolves. If Ollama is slow on the classify call (or hangs), the user can't interrupt.

Two options:
- Move `setStreaming(true)` (and pass the signal into `classifyAndDecide`) so Stop is live from the first millisecond.
- Document the gap and leave classify uncancellable since it's a single small call. Acceptable if classify is reliably fast — but worth confirming against the eval harness #140.

## Smaller

**3. `setStreaming(true)!` non-null assertion is brittle.**
```ts
const signal = this.setStreaming(true)!;
```
If a future branch returns `undefined` in `setStreaming(true)` (e.g. early-exit when already streaming), the `!` masks a null and you get a runtime `signal.aborted` on undefined. Two function overloads would let TS guarantee the return type:
```ts
private setStreaming(active: true): AbortSignal;
private setStreaming(active: false): void;
```

**4. Tool calls dispatched after the stream completes aren't cancellable.**
In `runChat`, once `services.llm.chat` resolves, the loop over `reply.toolCalls` calls `dispatchToolCall` without forwarding `signal`. If the LLM emits a `save_note` and the modal is open, Stop has no effect on the in-flight tool. Not a regression (this PR doesn't promise tool-call cancellation), but worth a follow-up issue when the dispatcher stack (#139/#141/#142) lands.

**5. Cosmetic — comment claim "Pop the assistant entry" is wrong even ignoring the bug above.**
The assistant entry is never pushed in the abort path; the line is either a no-op or (today) deletes the user message. Tighten the comment when fixing #1.

**6. Test coverage stops at `StreamingState`.**
The state machine is well-covered, but the view-level flow (click Stop → composer refilled → "stopped" tag rendered → next Send creates a new turnId) has no test. `src/view.test.ts` exists; even one happy-path test that mocks `services.llm.chat` to throw `AbortError` would catch bug #1.

## Praise (not blocking, just worth noting)

- `StreamingState` as a 47-line module is exactly the right size; the defensive `abort()` in `begin()` if a prior controller exists is correct and matches what I'd want.
- `isAbortError` exported and re-used aligns with the existing `classifier-llm` / `retry-policy` pattern.
- The single-handler button toggle (`if isStreaming → cancel, else → send`) is the cleanest way to wire this — no separate Stop button means no second focus-target.
- The cancelled-message styling (muted + italic + left border, distinct from error red) correctly signals "user-initiated stop" vs "failure".

Bug #1 is the blocker; everything else is polish.

heimdallSWE:
Thanks for the review — addressed in [4961eec](https://github.com/SimonProjekt/Gemmera/pull/145/commits/4961eec):

- **Bug #1 (blocker):** dropped the `this.history.pop()` in the AbortError branch of `runChat`. You're right — the assistant entry isn't pushed until `services.llm.chat` resolves, so on abort there's nothing to pop and the call was silently deleting the user message. Comment rewritten to match what the code actually does now (user message preserved, no assistant entry to pop, partial tokens stay in the DOM with the cancelled marker).
- **Polish #3 (typed overloads):** `setStreaming(active: true): AbortSignal` and `setStreaming(active: false): void`. Dropped the three `!` non-null assertions at the call sites.
- **Polish #5 (cosmetic):** comment tightened alongside the #1 fix.
- **Gap #2 (Stop during classify):** documented inline as a known limitation rather than wired in this PR. Threading signal through `classifyTurn` would touch the classifier-orchestrator public API and its test stack — that's separate-PR territory. Classify is a single small LLM call so the practical impact is small. Happy to open a follow-up issue if you'd rather have it tracked.
- **Note #4 (tool-call cancellation after stream):** acknowledged — out of scope for this PR, will revisit when the dispatcher stack (#139/#141/#142) settles.
- **Note #6 (view-level integration test):** not added in this iteration. The right harness for view tests doesn't exist yet (Obsidian DOM mocks). Worth its own issue.

715/715 still green; `tsc --noEmit` clean against this diff (the one remaining TS error in `note-preview-modal.ts` is from another in-flight branch).

iamaxeloberg (commented):
The fixes from the first review round look good — dropping the spurious `history.pop()`, adding typed overloads for `setStreaming`, and documenting the classify gap are all the right calls. A few remaining items:

## Bugs

**1. `runChat` missing `finally` to clear `inFlightText` on error.**

`runMixed` and `runAsk` both wrap their orchestrator calls in try/finally to clear `inFlightText`. `runChat` does not. If `services.llm.chat` throws a non-abort error, `inFlightText` leaks. The next send overwrites it, so this is low-severity, but it's inconsistent with the other two paths.

## Polish

**2. `cancelStreaming` does not clear `inFlightText`.**

After Stop, the composer is refilled and `inFlightText` still holds the old text. This is probably intentional (the user might want to edit-and-retry), but if `handleSend` fails early before reassigning `inFlightText`, the stale text from the prior cancelled turn persists. Consider adding `this.inFlightText = null` at the end of `cancelStreaming` or documenting the intentional retention.

**3. No view-level integration test.**

The `StreamingState` unit tests are thorough, but the view-level flow (click Stop → composer refilled → "stopped" tag rendered → next Send creates a new turnId) has no test. Even one happy-path test that mocks `services.llm.chat` to throw `AbortError` would have caught bug #1 from the first review round. Worth a follow-up issue to set up a minimal view test harness.

## Praise

- `StreamingState` as a 47-line module is exactly the right size. The defensive `abort()` in `begin()` if a prior controller exists is correct.
- Single-handler button toggle (Send ↔ Stop) is the cleanest approach — no second focus target.
- Cancelled-message styling (muted + italic + left border) correctly signals "user-initiated stop" vs "failure".

heimdallSWE:
Thanks @iamaxeloberg. Polish item #2 addressed in [2d6e796](https://github.com/SimonProjekt/Gemmera/pull/145/commits/2d6e796) — \`cancelStreaming\` now clears \`inFlightText\`. The textarea is the user-facing source of truth for the retry, so clearing the field eliminates the stale-value risk in any early-exit path through \`handleSend\`.

On item #1 (\`runChat\` missing \`finally\`): I think there's a stale read. The chat branch in \`handleSend\` already wraps \`runChat\` in:

\`\`\`ts
const signal = this.setStreaming(true);
try {
  await this.runChat(text, intent, turnId, route ?? null, signal);
} finally {
  this.setStreaming(false);
  this.inFlightText = null;
}
\`\`\`

…symmetric with the mixed and ask branches. If you're seeing a code path where \`inFlightText\` survives a non-abort error, happy to look — but reading current view.ts the bookkeeping looks consistent.

Item #3 (view-level integration test) — agreed, follow-up. The right move is a minimal Obsidian DOM mock + one happy-path test that throws AbortError from a mocked \`services.llm.chat\`. Will open an issue when this PR merges.

721/721, build clean.

PR #152, opened 2026-05-14
feat(chat): plugin-level chat state + inline rename + retention setting (#43)

### LINKED ISSUE #43: Plugin-level chat state and chats.duckdb persistence
Conversation state lives on the plugin instance, not on the leaf, so popping between sidebar/tab/pop-out preserves the live chat. Persist chat history to `vault/.coworkmd/chats.duckdb` — separate from the index.

Schema per chat: stable ID, auto-generated title (editable, derived from first user message), created timestamp, ordered list of turns. "History" drawer inside the view lists prior chats.

Default retention: unlimited. Settings option to cap at N days or N chats (UI lives in Settings tab issue).

### Acceptance
- Chats survive plugin reload and Obsidian restart.
- Detaching the view to a pop-out window keeps the conversation live.
- History drawer opens, lists chats, and lets the user load a prior chat.
- Title auto-generates from first user message and is editable inline.
- Deleting `chats.duckdb` does not affect the RAG index file.

### Reference
`planning/ui-surfaces.md` — "State and persistence".

heimdallSWE:
Closes #43. PR 3 of the UI-surfaces v1 sequence (after #144 / #145).

Builds on top of what #70 already shipped (history drawer, auto-titling, retention plumbing) and closes the three remaining acceptance gaps for #43.

### Pop-out continuity
- New `ChatSessionState` (`src/services/chat-session-state.ts`) lives on the plugin instance and owns the in-memory \`history\`, \`currentSessionId\`, and \`recentTurns\`.
- \`GemmeraChatView\` takes it as a constructor injection and exposes it via getter/setter shims so existing view code reads/writes shared state with no call-site refactoring.
- \`ChatHistoryStore\` is also now plugin-owned (lazy constructor in \`getChatHistory()\`) so all view instances share one persistence handle.
- \`onOpen\` replays buffered turns into \`messagesEl\`, so detaching the view to a pop-out window picks up the live conversation instead of resetting.

### Inline title editing
- New \`ChatHistoryStore.renameSession(id, title)\` — trims, rejects empty, clamps to 200 chars, bumps \`updatedAt\` so the renamed chat surfaces as recent. **5 new unit tests** cover happy / empty / long / unknown / ordering.
- Drawer titles get a double-click → contenteditable handler. Enter saves, Escape reverts, blur saves. Click propagation stops so editing doesn't switch sessions mid-rename.

### Retention setting
- Added \`chatRetentionMaxDays\` + \`chatRetentionMaxSessions\` to \`GemmeraSettings\` (default 0 = unlimited).
- Two new entries in Settings → tab, wired to the existing \`pruneIfNeeded\` path that #70 set up.

## Acceptance check (issue #43)

| Criterion | Status |
|---|---|
| Chats survive plugin reload + Obsidian restart | ✅ (already shipped via \`chats.json\`) |
| Detaching the view to pop-out keeps conversation live | ✅ Plugin-level \`ChatSessionState\` + \`onOpen\` replay |
| History drawer opens, lists, loads a prior chat | ✅ (already shipped in #70) |
| Title auto-generates from first user message and is editable inline | ✅ Auto from #70, inline edit new in this PR |
| Deleting \`chats.duckdb\` (we use \`chats.json\`) doesn't affect the RAG index file | ✅ Separate file under \`.coworkmd/\` |

## Implementation notes

- Memory record \`rag_design_decisions\` says JSON + binary storage for now, no DuckDB yet. The issue title mentions \`chats.duckdb\` but the existing implementation uses \`chats.json\`; spec compliance is via the **schema** (sessions list with stable id + title + turns + timestamps) and the **API shape** (\`createSession\`, \`loadSession\`, \`appendTurn\`, \`listSessions\`, \`renameSession\`, \`pruneIfNeeded\`) which are DuckDB-friendly and let us swap the backing store without touching the view.
- One pre-existing \`tsc\` error in \`note-preview-modal.ts\` from another in-flight PR — unrelated to this diff.

## Test plan

- [x] \`npm test\` → 716 / 716 (up from 715)
- [x] \`npm run build\` → clean
- [ ] Manual: open chat → send a turn → detach to new window → live conversation appears
- [ ] Manual: double-click a chat title in the drawer → rename → Enter persists → reopen → new title sticks
- [ ] Manual: set max sessions = 2 → start a third chat → next \`onOpen\` prunes the oldest

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -305,17 +305,18 @@ export default class GemmeraPlugin extends Plugin {
 
   /**
    * Lazy accessor for the shared chat-history store. Constructed on first
-   * call once the vault adapter is known. Retention policy reads from the
-   * plugin settings so toggling them takes effect on the next prune. #43.
+   * call once the vault adapter is known. Passes a *getter* for the retention
+   * policy so live edits to the settings sliders take effect on the next
+   * prune without a plugin reload (#152 review).
    */
   private getChatHistory(): ChatHistoryStore {
     if (!this.chatHistory) {
       const adapter = this.app.vault.adapter as FileSystemAdapter;
       const historyPath = adapter.getFullPath(".coworkmd/chats.json");
-      this.chatHistory = new ChatHistoryStore(historyPath, {
+      this.chatHistory = new ChatHistoryStore(historyPath, () => ({
         maxDays: this.settings.chatRetentionMaxDays > 0 ? this.settings.chatRetentionMaxDays : undefined,
         maxSessions: this.settings.chatRetentionMaxSessions > 0 ? this.settings.chatRetentionMaxSessions : undefined,
-      });
+      }));
     }
     return this.chatHistory;
   }
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,6 +2,8 @@ import { FileSystemAdapter, Menu, Notice, Plugin, TFile, WorkspaceLeaf } from "o
 import { readFileSync, existsSync, watch as fsWatch } from "fs";
 import { GemmeraChatView, VIEW_TYPE } from "./view";
 import { createServices, Services } from "./services";
+import { ChatHistoryStore } from "./services/chat-history";
+import { ChatSessionState } from "./services/chat-session-state";
 import { OllamaLifecycle, type OllamaStatus } from "./services/ollama-lifecycle";
 import { DEFAULT_SETTINGS, GemmeraSettings } from "./settings";
 import { GemmeraSettingsTab } from "./ui/settings-tab";
@@ -14,6 +16,10 @@ export default class GemmeraPlugin extends Plugin {
   private statusBarEl!: HTMLElement;
   private batteryTimer: ReturnType<typeof setInterval> | null = null;
   private batteryPaused = false;
+  // Plugin-instance-level chat state so pop-out / re-open survives (#43).
+  // Constructed lazily on first view open because we need the vault adapter.
+  private chatHistory: ChatHistoryStore | null = null;
+  private chatSessionState = new ChatSessionState();
 
   async onload(): Promise<void> {
     this.settings = Object.assign({}, DEFAULT_SETTINGS, await this.loadData());
@@ -75,7 +81,13 @@ export default class GemmeraPlugin extends Plugin {
 
     this.registerView(
       VIEW_TYPE,
-      (leaf) => new GemmeraChatView(leaf, this.services, this.settings),
+      (leaf) => new GemmeraChatView(
+        leaf,
+        this.services,
+        this.settings,
+        this.getChatHistory(),
+        this.chatSessionState,
+      ),
     );
 
     const ribbonEl = this.addRibbonIcon("message-square", "Gemmera", () => {
@@ -289,6 +301,23 @@ export default class GemmeraPlugin extends Plugin {
       const count = "linkCount" in e ? e.linkCount : 0;
       console.debug(`[gemmera] links:${e.kind} ${e.path} (${count})`);
     });
+  }
+
+  /**
+   * Lazy accessor for the shared chat-history store. Constructed on first
+   * call once the vault adapter is known. Retention policy reads from the
+   * plugin settings so toggling them takes effect on the next prune. #43.
+   */
+  private getChatHistory(): ChatHistoryStore {
+    if (!this.chatHistory) {
+      const adapter = this.app.vault.adapter as FileSystemAdapter;
+      const historyPath = adapter.getFullPath(".coworkmd/chats.json");
+      this.chatHistory = new ChatHistoryStore(historyPath, {
+        maxDays: this.settings.chatRetentionMaxDays > 0 ? this.settings.chatRetentionMaxDays : undefined,
+        maxSessions: this.settings.chatRetentionMaxSessions > 0 ? this.settings.chatRetentionMaxSessions : undefined,
+      });
+    }
+    return this.chatHistory;
   }
 
   private async openChatView(): Promise<void> {
--- a/src/services/chat-history.test.ts
+++ b/src/services/chat-history.test.ts
@@ -127,4 +127,52 @@ describe("ChatHistoryStore", () => {
     const loaded = await store.loadSession(session.id);
     expect(loaded).not.toBeNull();
   });
+
+  // ── renameSession (#43) ────────────────────────────────────────────────
+
+  it("renameSession updates the title and persists it", async () => {
+    const store = new ChatHistoryStore(storePath());
+    const session = await store.createSession();
+    const updated = await store.renameSession(session.id, "Sundsvall trip");
+    expect(updated?.title).toBe("Sundsvall trip");
+    const reloaded = await store.loadSession(session.id);
+    expect(reloaded?.title).toBe("Sundsvall trip");
+  });
+
+  it("renameSession trims whitespace and clamps to 200 chars", async () => {
+    const store = new ChatHistoryStore(storePath());
+    const session = await store.createSession();
+    const long = "x".repeat(300);
+    const updated = await store.renameSession(session.id, `   ${long}   `);
+    expect(updated?.title).toHaveLength(200);
+  });
+
+  it("renameSession rejects an empty / whitespace title", async () => {
+    const store = new ChatHistoryStore(storePath());
+    const session = await store.createSession();
+    expect(await store.renameSession(session.id, "")).toBeNull();
+    expect(await store.renameSession(session.id, "   ")).toBeNull();
+    const reloaded = await store.loadSession(session.id);
+    expect(reloaded?.title).toBe("New chat");
+  });
+
+  it("renameSession returns null for unknown id", async () => {
+    const store = new ChatHistoryStore(storePath());
+    const result = await store.renameSession("does-not-exist", "Anything");
+    expect(result).toBeNull();
+  });
+
+  it("renameSession bumps updatedAt so the renamed chat surfaces as recent", async () => {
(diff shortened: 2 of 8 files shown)


SimonProjekt (changes requested):
I found one blocking issue before this should merge.

`GemmeraPlugin.getChatHistory()` builds `ChatHistoryStore` only once and passes the current retention values into the constructor. After the chat view has been opened once, changing `chatRetentionMaxDays` or `chatRetentionMaxSessions` in Settings only mutates `this.settings`; the cached `this.chatHistory` keeps the old retention object forever. That means the new retention controls usually do not take effect on the next chat open as the Settings UI promises, and if the first open happened with the default `0` values, pruning remains unlimited until the plugin is reloaded.

Relevant code:
- `src/main.ts`: `getChatHistory()` returns the cached store after first construction.
- `src/ui/settings-tab.ts`: the setting handlers save new values but do not update or recreate the store.
- `src/services/chat-history.ts`: `pruneIfNeeded()` reads the constructor-provided retention policy, not the live plugin settings.

Please make the retention policy live after settings changes. A small fix would be to either recreate/update the cached `ChatHistoryStore` when retention settings change, or let `ChatHistoryStore` receive a policy getter / setter so `pruneIfNeeded()` reads the current values.

- `git diff --check origin/dev...origin/pr-152`
- `npm test` — 716 tests passed
- `npm run build`

heimdallSWE (commented):
Solid PR — `ChatSessionState` as a tiny plugin-level class is the right abstraction, and the getter/setter shims avoid touching every call site. The `renameSession` test coverage (5 cases) is excellent. One concrete bug, a real concurrency concern that was latent in #70 but gets worse here, and a handful of smaller items.

## Bug

**1. Retention setting cache: the lazy accessor never rebuilds, but the UI claims it does.**

In `main.ts`:
```ts
private getChatHistory(): ChatHistoryStore {
  if (!this.chatHistory) {
    ...
    this.chatHistory = new ChatHistoryStore(historyPath, {
      maxDays: this.settings.chatRetentionMaxDays > 0 ? ... : undefined,
      maxSessions: this.settings.chatRetentionMaxSessions > 0 ? ... : undefined,
    });
  }
  return this.chatHistory;
}
```

`ChatHistoryStore` captures the retention object by reference once and never re-reads. But the settings-tab description says:
> "Takes effect on the next chat open."

Once `chatHistory` is constructed, opening another chat returns the cached instance with the old retention. Real behavior is "takes effect after plugin reload." Three options:

- Cheapest: change the description to "Takes effect after plugin reload."
- Better: when retention settings change, null-out `this.chatHistory` so the next `getChatHistory()` rebuilds.
- Best: make `ChatHistoryStore` read retention via a callback `() => GemmeraSettings`, so it always sees current values.

## Concurrency

**2. `renameSession` + `appendTurn` race-write the JSON file.**

Both follow read → mutate → flush. With shared `chatHistory` across all view instances (sidebar / tab / pop-out), the chance of a concurrent rename + turn-append (or two turn-appends from different views) goes up. `flush` ends with an atomic `rename`, but the read-modify cycle isn't serialized — the loser's mutation gets silently dropped.

This is pre-existing from #70, not introduced here, but this PR materially widens the surface (multiple live views now share one store). Worth at least a follow-up issue: serialize all writes via an in-memory mutex/queue, or move to a single mutating actor.

## Pop-out / replay

**3. `onOpen` replay loses message decorations.**

```ts
for (const turn of this.history) {
  this.appendMessage(turn.role === "user" ? "user" : "assistant", turn.content);
}
```

`appendMessage(role, content, decoration?)` accepts a decoration arg (route rationale, silentSave indicator, etc.). The pop-out replay passes only role + content, so route chips and inline classifier hints disappear when the view is detached. The conversation text survives — which satisfies the AC literally — but the pop-out has visibly less context than the original. Either:

- Store decorations alongside each turn in `ChatSessionState`, or
- Accept that pop-out is "text-only" and document.

**4. Replay also doesn't restore citation chips.**

Same root cause: `extractWikilinks(reply.content)` ran on the original assistant turn and was added to the DOM; the replay just re-renders the markdown text. Wikilink chips disappear after pop-out. Probably acceptable v1.

## Smaller

**5. Inline rename: `keydown` listener leaks per edit session.**

```ts
titleEl.addEventListener("blur", () => { void commit(); }, { once: true });
titleEl.addEventListener("keydown", (e) => { ... });  // not once
```

The blur listener self-cleans (`once: true`), but the keydown listener stays attached to the title element until `renderHistoryDrawer` rebuilds the DOM. Cancel-then-rename-again on the same node accumulates listeners. `renderHistoryDrawer` runs after a successful commit, so in practice it gets garbage-collected — but if the user double-clicks → Escape → double-clicks again, you have two keydown handlers on a still-attached node, both firing. Wrap with an `AbortController`:

```ts
const ac = new AbortController();
titleEl.addEventListener("blur", () => { ac.abort(); void commit(); }, { once: true });
titleEl.addEventListener("keydown", handler, { signal: ac.signal });
```

**6. Ad-hoc `TextComponent` typing in settings-tab.ts.**

```ts
.addText((text: { setValue: (v: string) => unknown; onChange: (cb: (v: string) => void) => unknown; setPlaceholder?: (s: string) => unknown }) => { ... })
```

Obsidian exports `TextComponent`. The structural-type workaround is harder to read than `(text: TextComponent) => void` and breaks the moment you need another method on it. Worth aligning with how the rest of the settings-tab types component callbacks.

**7. `chatRetentionMaxDays` / `MaxSessions` silently coerce invalid input to 0.**

`parseInt("abc")` → NaN → 0 (unlimited). User typing a typo sets retention to "unlimited" without feedback. Either add a validation hint (red border + "Must be a non-negative integer") or restrict the input via `type="number" min="0"`. Low priority.

**8. `renameSession` bumps `updatedAt` so renames surface to the top.**

The test asserts this intentionally. Worth surfacing in the inline doc — users may rename old chats to organize them, expecting position to stay; this UI re-orders. Not wrong, just opinionated.

## Praise

- **`ChatSessionState.reset()` fixes the `recentTurns`-not-cleared bug** I flagged on #143. Nice incidental win — the classifier no longer leaks intent across new-chat boundaries.
- **Getter/setter shims keep the diff to view.ts small** while moving state ownership out. Clean migration tactic.
- **`renameSession` testing is thorough** — empty/whitespace, length clamp, unknown id, ordering. Sets the right bar for the rest of the stack.
- **`pruneIfNeeded` no-op short-circuit (from #143 fix) is preserved**; new retention settings cleanly compose with it.

Concrete blockers: #1 (cache vs. description mismatch — pick a fix). The rest is polish or follow-ups.

heimdallSWE:
Thanks for the catch — addressed in [b9697bf](https://github.com/SimonProjekt/Gemmera/pull/152/commits/b9697bf).

Going with your second suggestion (policy getter) rather than recreating the store, since it keeps lifecycle simple and avoids any \`onChange\` plumbing in the settings tab:

- \`ChatHistoryStore\` now accepts \`ChatRetentionSource = ChatRetentionPolicy | (() => ChatRetentionPolicy)\`. A private getter unwraps it on each call.
- \`pruneIfNeeded()\` reads through the getter, so live edits to either slider take effect on the next prune (next chat open) with no reload.
- \`main.ts\` constructs the store with a getter that closes over \`this.settings\`, so settings-tab handlers only need to mutate the field.
- Added a regression test: flip \`maxSessions\` between two \`pruneIfNeeded()\` calls and confirm the second prune respects the new value.
- Settings descriptions tightened from "Takes effect on the next chat open" → "Takes effect on the next prune (chat open)" so the contract matches what \`pruneIfNeeded\` actually does.

717/717 tests (added 1), build clean, CI green.

iamaxeloberg (commented):
The retention getter fix is clean — passing a closure over `this.settings` is the right call and the regression test is solid. A few remaining items after the fix:

## Bugs

**1. Inline rename keydown listener leak.**

In `beginInlineRename` (`view.ts`):
```ts
titleEl.addEventListener("blur", () => { void commit(); }, { once: true });
titleEl.addEventListener("keydown", (e) => { ... });  // no cleanup
```

The blur listener self-cleans via `{ once: true }`, but the keydown listener stays attached to the DOM node. If the user double-clicks → Escape → double-clicks again on the same element (before `renderHistoryDrawer` rebuilds the DOM), two keydown handlers fire simultaneously. Wrap both in an `AbortController`:

```ts
const ac = new AbortController();
titleEl.addEventListener("blur", () => { ac.abort(); void commit(); }, { once: true });
titleEl.addEventListener("keydown", handler, { signal: ac.signal });
```

## Polish

**2. Ad-hoc `TextComponent` typing in settings-tab.ts.**

```ts
.addText((text: { setValue: (v: string) => unknown; onChange: ... }) => { ... })
```

Obsidian exports `TextComponent`. The structural-type workaround is harder to read and breaks the moment you need another method on the component. Use `(text: TextComponent) => void` like the rest of the settings tab does.

**3. Retention inputs silently coerce invalid input to 0 (unlimited).**

`parseInt("abc")` → NaN → 0. A user typo sets retention to "unlimited" with no feedback. Either add a validation hint or switch to `type="number"` with `min="0"`. Low priority but easy win for UX.

**4. `renameSession` bumps `updatedAt` so renamed chats jump to the top.**

The test asserts this intentionally. Worth a one-line doc comment so future maintainers know this is deliberate — users may rename old chats to organize them and expect position to stay.

## Praise

- `ChatSessionState.reset()` fixing the `recentTurns`-not-cleared bug from #143 is a nice incidental win.
- Getter/setter shims keep the diff to view.ts small while moving state ownership out. Clean migration.
- `renameSession` test coverage (5 cases) sets the right bar.

PR #153, opened 2026-05-14
feat(ui): NotePreviewModal frontmatter form + commit gate (#52)

### LINKED ISSUE #52: NotePreviewModal — pre-save commit gate
`NotePreviewModal` is a dedicated `Modal` subclass shown before any note is written to the vault. Appears regardless of sidebar/tab mode because it is a commit gate, not a conversation surface.

Contents:
- Editable title.
- Editable folder path (defaults to `Inbox/`, configurable in settings).
- Editable frontmatter rendered as a form over the JSON contract from `rag.md`, with type-aware widgets for tags, aliases, status.
- Markdown body preview via `MarkdownRenderer.renderMarkdown` so it matches Obsidian's native rendering.
- "Split into multiple notes" toggle: when enabled, lists proposed notes and lets user confirm each independently.
- Confirm / Edit / Cancel buttons.

### Acceptance
- Modal blocks file write until confirmed.
- Frontmatter editor enforces the `rag.md` schema (rejects invalid types).
- Body preview is visually identical to native rendering.
- Split mode produces N independent confirm flows; cancelling one does not cancel the others.
- Keyboard: `Tab` cycles fields, `Ctrl/Cmd+Enter` confirms, `Esc` cancels.

### Reference
`planning/ui-surfaces.md` — "Preview modal (pre-save confirmation)". Cross-ref `planning/rag.md` for frontmatter contract.

heimdallSWE:
Closes #52 (modal commit gate scope). Split-mode AC tracked separately in #160.

PR 3 of the UI-surfaces v1 sequence (PR 1 = #144, PR 2 = #145).

## Summary

- `NotePreviewModal` becomes a real commit gate: editable title, folder, **type**, **status**, **tags**, **aliases**, and **summary** over the rag.md frontmatter contract.
- `dispatchSaveCreate` builds a complete `NoteSpec` and routes through `composeFile` (the same serializer the ingest pipeline uses) — guarantees every required schema key is emitted, including `source`, `entities`, `related`, `summary`, `key_points`, and the nested `cowork` object. Empty arrays render as `[]`.
- Schema validation rejects empty/over-long titles, out-of-enum type/status, and empty/over-long summaries with an inline error.
- `updateSaveEnabled` runs the full validator so the Save button stays gated even if type/status `<select>` elements are later swapped for free-text inputs.
- `Cmd/Ctrl+Enter` confirms from anywhere in the modal; Esc and outside-click still cancel.
- `validateNotePreview` is unit-tested in isolation (no DOM); covers all reject paths plus CSV trimming and folder fallback.

## Acceptance (issue #52)

| Criterion | Status |
|---|---|
| Modal blocks file write until confirmed | ✅ Promise-based gate; `null` on cancel |
| Frontmatter editor enforces rag.md schema | ✅ `validateNotePreview` rejects bad title/type/status/summary; selects restrict at input level; written frontmatter now passes the full schema |
| Body preview matches native rendering | ✅ Existing `MarkdownRenderer.render` path retained |
| Keyboard: Tab cycles, Cmd/Ctrl+Enter confirms, Esc cancels | ✅ Tab is the browser default; `Scope.register(["Mod"], "Enter")` for confirm; `Modal` handles Esc |
| Split into multiple notes | ⏭ Tracked separately in **#160** — no upstream caller passes >1 spec today; will land alongside `propose_notes` |

## Review fixes applied

**SimonProjekt's review (CHANGES_REQUESTED):**
1. Schema gap — `dispatchSaveCreate` now emits all 12 required keys via `composeFile` (canonical serializer). Empty `tags` / `aliases` / `entities` / `related` render as `[]`, never omitted.
2. Split-mode AC — moved to follow-up issue #160; PR scope explicitly narrowed to "modal commit gate."

**iamaxeloberg's review (COMMENTED):**
3. `updateSaveEnabled` now runs the full validator instead of gating only on title.
4. `.gemmera-note-preview__error` falls back to a readable red (`var(--text-error, #e74c3c)`) for themes that don't define the variable.
5. CSV quoting in `splitCsv` — deferred (low priority; users rarely put commas in tags).

iamaxeloberg's items 1 and 2 reference `runChat` / `cancelStreaming` / `inFlightText` in `view.ts`, which this PR doesn't touch — they belong to #145's surface and should be a follow-up there if the concerns are real.

## Test plan

- [x] `npm test` → 734/734 passing (was 722; +12 schema-compliance + summary validation tests)
- [x] `npm run build` → clean
- [x] `tsc --noEmit` → only the pre-existing `MarkdownRenderer.render(..., this)` typing nit on dev; no new errors

### DIFF
--- a/src/services/ingest-writer.ts
+++ b/src/services/ingest-writer.ts
@@ -109,7 +109,7 @@ async function uniquePath(
  * its own frontmatter block (the parser strips it; we re-fail here as
  * defense in depth so a buggy parser cannot corrupt the file format).
  */
-function composeFile(spec: NoteSpec): string {
+export function composeFile(spec: NoteSpec): string {
   const body = spec.body_markdown ?? "";
   if (/^---\s*\r?\n/.test(body)) {
     throw new Error("body_markdown must not contain its own frontmatter block");
--- a/src/services/tool-dispatcher.test.ts
+++ b/src/services/tool-dispatcher.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 import { MockVaultService } from "../contracts/mocks/mock-vault";
-import { dispatchToolCall, buildFrontmatter, appendUnderDatedSection, patchFrontmatter, type ToolDispatchDeps } from "./tool-dispatcher";
+import { dispatchToolCall, appendUnderDatedSection, patchFrontmatter, type ToolDispatchDeps } from "./tool-dispatcher";
 import { InMemoryJobQueue } from "./in-memory-job-queue";
 import { IngestWriter } from "./ingest-writer";
 
@@ -22,21 +22,23 @@ function makeDeps(overrides: Partial<ToolDispatchDeps> = {}): ToolDispatchDeps {
   };
 }
 
+const fullConfirm = {
+  confirmed: true as const,
+  title: "Dogs",
+  folder: "Inbox/",
+  tags: [],
+  aliases: [],
+  type: "source" as const,
+  status: "inbox" as const,
+  summary: "Notes on dogs.",
+};
+
 describe("dispatchToolCall — routing", () => {
   it("routes save_note with mode=create to save handler", async () => {
-    const confirmResult = {
-      confirmed: true as const,
-      title: "Dogs",
-      folder: "Inbox/",
-      tags: [],
-      aliases: [],
-      type: "source" as const,
-      status: "inbox" as const,
-    };
-    const deps = makeDeps({ openNotePreview: vi.fn().mockResolvedValue(confirmResult) });
+    const deps = makeDeps({ openNotePreview: vi.fn().mockResolvedValue(fullConfirm) });
 
     const result = await dispatchToolCall(
-      { id: "1", name: "save_note", arguments: { mode: "create", title: "Dogs", body_markdown: "Dogs are great." } },
+      { id: "1", name: "save_note", arguments: { mode: "create", title: "Dogs", body_markdown: "Dogs are great.", summary: "Notes on dogs." } },
       deps,
     );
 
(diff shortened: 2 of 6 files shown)


SimonProjekt (changes requested):
I found two blocking acceptance issues before this should merge.

1. The create path still does not write notes that satisfy the `rag.md` frontmatter contract. `dispatchSaveCreate()` now passes `title`, `type`, `status`, `tags`, and `aliases` into `buildFrontmatter()`, but the schema requires additional fields such as `source`, `entities`, `related`, `summary`, `key_points`, and `cowork`. It also requires `tags` and `aliases` even when they are empty, while `buildFrontmatter()` omits those keys for empty arrays. As a result, a confirmed modal save can still produce frontmatter that fails `schemas/frontmatter.schema.json`, even though issue #52 requires the frontmatter editor to enforce the `rag.md` schema.

Relevant code:
- `src/services/tool-dispatcher.ts`: `dispatchSaveCreate()` only forwards the modal subset to `buildFrontmatter()`.
- `src/services/tool-dispatcher.ts`: `buildFrontmatter()` only emits the subset and omits empty `tags` / `aliases`.
- `src/ui/note-preview-modal.ts`: `validateNotePreview()` validates only title/type/status plus CSV parsing, not the full required schema object.

2. The PR closes #52 but does not implement the split-mode acceptance criterion. The issue explicitly requires: “Split mode produces N independent confirm flows; cancelling one does not cancel the others.” The PR description says this is deferred, which may be a reasonable product decision, but then this PR should not close #52 as-is. Either implement split mode here or adjust the issue/PR scope so the remaining acceptance criterion is tracked separately before this merges.

- `git diff --check origin/dev...origin/pr-153`
- `npm test` — 722 tests passed
- `npm run build`
- `npx tsc --noEmit` — fails only on the pre-existing `MarkdownRenderer.render(..., this)` type issue that is already present on `dev`

iamaxeloberg (commented):
Good expansion of the modal into a full commit gate. A few additional findings on top of SimonProjekt's review:

## Bugs

**1. `runChat` missing `finally` to clear `inFlightText` on error path.**

In `view.ts`, `runMixed` and `runAsk` both wrap their orchestrator calls in try/finally to clear `inFlightText`. `runChat` does not. If `services.llm.chat` throws a non-abort error, `inFlightText` is leaked. The next send overwrites it, so this is low-severity, but it's inconsistent and could cause confusion during debugging.

**2. `cancelStreaming` does not clear `inFlightText`.**

After Stop, the composer is refilled and `inFlightText` still holds the old text. If the user clicks Send without editing, the same text goes out as a new turn — which is fine — but if `handleSend` fails early (before `this.inFlightText = text` at the top), the stale text from the prior cancelled turn would be used. Worth adding `this.inFlightText = null` at the end of `cancelStreaming` or documenting the intentional retention.

## Polish

**3. `updateSaveEnabled` only gates on title.**

The Save button is disabled when the title is empty, but nothing disables it when type/status are invalid. In practice the `<select>` elements restrict input to valid enum values, so this is currently safe. But if someone later swaps a select for a free-text input, the gate would be incomplete. Consider having `updateSaveEnabled` also call `validateNotePreview` and disable on any error.

**4. `--text-error` CSS variable may not exist in all Obsidian themes.**

The new `.gemmera-note-preview__error` rule uses `var(--text-error)`. This isn't a standard Obsidian CSS variable — some themes define it, others don't. If undefined, the error text will be invisible. Consider falling back to a standard variable like `var(--text-accent-failed)` or adding a hardcoded fallback color.

**5. `splitCsv` doesn't handle quoted CSV values.**

If a user types `"work, important", personal` as tags (intending a tag with a comma inside), `splitCsv` will produce three entries instead of two. Low priority since tags with commas are unusual, but worth noting if rag.md expects proper CSV parsing.

heimdallSWE:
Pushed `890a7b8` addressing the reviews.

**@SimonProjekt** — both blockers addressed:

1. **Schema compliance** — `dispatchSaveCreate` now builds a complete `NoteSpec` and routes through `composeFile` (extracted from `IngestWriter`, same path synthesis notes use). Every required key from `frontmatter.schema.json` is emitted, including `source`, `entities`, `related`, `summary`, `key_points`, and the nested `cowork` block. Empty arrays render as `[]` rather than being omitted. New tests in `tool-dispatcher.test.ts` lock this down (schema-compliance + `cowork.model`/`run_id` stamping). The partial `buildFrontmatter` helper is removed — no other callers.

2. **Split mode** — moved to follow-up issue #160. PR description updated from `Closes #52` to `Refs #52`. #52 stays open until #160 lands; the modal portion is shipped here.

**@iamaxeloberg** — thanks for the review.

- **#3 (updateSaveEnabled)**: applied. Save button now disabled whenever `validateNotePreview` returns an error, including on type/status/summary/folder, so the gate stays correct even if a `<select>` is later swapped for a free-text input.
- **#4 (`--text-error` fallback)**: applied. `color: var(--text-error, #e74c3c);` keeps the error message visible in themes that don't define the variable.
- **#5 (`splitCsv` quoting)**: deferring — tags with embedded commas are an edge case and worth a separate issue if anyone hits it.
- **#1 and #2** (`runChat` `finally`, `cancelStreaming` clearing `inFlightText`): these reference `view.ts` code that #153 doesn't touch — that surface was introduced by #145. If the concerns are real, they belong as a follow-up against post-#145 dev. Happy to file an issue if you'd like.

`summary` is now a required frontmatter field per the rag.md schema (1–600 chars). Added as a textarea in the modal and as an optional arg on the `save_note` tool so the LLM can pre-fill but the user can always edit. Rejected empty/over-long with the same inline error path used by title/type/status.

Tests: 734/734 passing. tsc clean modulo the pre-existing `MarkdownRenderer.render(..., this)` issue already on `dev`.

Ready for re-review.

PR #154, opened 2026-05-14
fix(build): add typecheck to CI

### LINKED ISSUE #146: fix(build): add TypeScript typecheck to CI
## Problem

`npm run build` currently passes because esbuild transpiles without full TypeScript checking, but a standalone typecheck fails:

```text
src/ui/note-preview-modal.ts(76,70): error TS2345: Argument of type 'this' is not assignable to parameter of type 'Component'.
```

This means CI can miss type-level regressions even when tests and production bundling pass.

## Acceptance criteria

- Add an npm script such as `typecheck` that runs `tsc --noEmit`.
- Fix the `MarkdownRenderer.render(...)` owner argument in `NotePreviewModal` using a type-correct Obsidian `Component` owner or supported rendering API.
- Run typecheck in CI alongside tests and production build.
- `npm test`, `npm run build`, and `npm run typecheck` all pass locally.

## Notes

Found during project-wide audit after PRs #141, #142, and #143 merged.

SimonProjekt:
## What

- Adds `npm run typecheck` as a dedicated `tsc --noEmit` check.
- Runs typecheck in CI before tests.
- Fixes `NotePreviewModal` Markdown rendering by passing a real Obsidian `Component` owner to `MarkdownRenderer.render` and unloading it with the modal.
- Updates the Obsidian test mock for `Component` and `MarkdownRenderer.render`.

## Why

`npm run build` uses esbuild and can pass while TypeScript errors remain hidden. The current failure was in the note preview modal's Markdown renderer owner argument.

Closes #146

## How to test

- `npm run typecheck`
- `npm test`
- `npm run build`
- `npm run eval:retrieval:ci`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,9 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
+      - name: Run typecheck
+        run: npm run typecheck
+
       - name: Run tests
         run: npm test
 
--- a/__mocks__/obsidian.ts
+++ b/__mocks__/obsidian.ts
@@ -11,7 +11,13 @@ export class FileSystemAdapter {
 export class Plugin {}
 export class WorkspaceLeaf {}
 export class ItemView {}
-export class MarkdownRenderer {}
+export class Component {
+  load(): void {}
+  unload(): void {}
+}
+export class MarkdownRenderer {
+  static async render(): Promise<void> {}
+}
 
 // Minimal stand-ins for tests that reference UI types but don't actually
 // render. Real Obsidian provides full implementations at runtime.
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "node esbuild.config.mjs",
     "build": "node esbuild.config.mjs production",
+    "typecheck": "tsc --noEmit",
     "test": "vitest run",
     "test:watch": "vitest",
     "eval:classifier": "tsx eval/runner/classify.ts",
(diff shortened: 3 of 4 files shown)


heimdallSWE (commented):
Small, targeted fix — adds `tsc --noEmit` to CI and resolves the long-standing `MarkdownRenderer.render(..., this)` typing issue that's been called out in three separate PR reviews. LGTM with a couple of small notes.

## Praise

- **Owning `Component` lifecycle explicitly.** Creating a `renderOwner` field, calling `load()` in `onOpen` and `unload()` in `onClose` is the right Obsidian pattern. The modal class doesn't extend `Component` itself (it extends `Modal`), so passing `this` to `MarkdownRenderer.render` was always a type lie — this fixes both the type error and the actual lifecycle semantics.
- **Mock additions are minimal.** `Component` and `MarkdownRenderer.render` stubs in `__mocks__/obsidian.ts` are scoped exactly to what tests need; no over-mocking.

## Smaller

- **Mock `Component` is structurally narrow.** `class Component { load(): void {} unload(): void {} }` is sufficient for the modal's two call sites, but real Obsidian `Component` also has `onload`, `onunload`, `addChild`, `removeChild`, `register`, `registerEvent`, etc. If a future test exercises any of these via a mocked Component, the test will fail with a runtime undefined-call. Not blocking — call out in a TODO comment if you want, or just wait for the first test that needs it.
- **`renderOwner.load()` runs even in branches where `MarkdownRenderer.render` throws synchronously.** Unlikely in practice — `render` returns a Promise and the `.catch(() => ...)` handles failure — but if `render` ever rejected synchronously (e.g. an arg-validation throw), `load()` would have run with no matching `unload()` if the modal stayed unopened. Edge case; safe in practice.
- **`typecheck` script is just `tsc --noEmit`.** No mention of `--incremental` or a cache dir. CI cold-starts will recompile from scratch each run — fine for now at the project's size; revisit if CI gets noticeably slower.

LGTM. Want me to also add a quick CI badge for the typecheck job to the README in a follow-up, or is that overkill?

PR #156, opened 2026-05-14
fix(classifier): improve mixed and meta routing

### LINKED ISSUE #150: fix(classifier): improve mixed and meta routing in golden eval
## Problem

The classifier CI eval passes the regression gate, but current mock accuracy is only 24/30 (80%). The remaining failures are concentrated in mixed and meta cases:

- `m002`: expected mixed, got capture
- `m004`: expected mixed, got capture
- `m006`: expected mixed, got ask
- `x003`: expected meta, got ask
- `x004`: expected meta, got capture
- `e001`: expected meta, got ask

These are exactly the ambiguous cases that affect routing and tool selection.

## Acceptance criteria

- Improve classifier routing for mixed save+ask turns and app-meta questions.
- Keep the P0 regression gate passing.
- Update prompts, heuristics, or golden expectations only when behavior is intentional and documented.
- `npm run eval:classifier:ci` passes with improved accuracy or a documented baseline update.
- `npm test` passes.

SimonProjekt:
## What
- Tightens mock classifier routing for mixed save-and-retrieve turns and meta/help feature questions.
- Adds prompt rules and few-shot examples for mixed/meta edge cases.
- Updates the classifier prompt snapshot.

## Why
Current classifier eval coverage had failures concentrated in mixed and meta routes. This brings the mock eval route behavior back in line with the golden set.

Closes #150

## How to test
- npm run eval:classifier:ci
- npm test
- npm run build

### DIFF
--- a/eval/runner/classify.ts
+++ b/eval/runner/classify.ts
@@ -89,18 +89,18 @@ class ScriptedClassifierLLM implements LLMService {
     let confidence = 0.78;
     let rationale = "Default: treating as a question.";
 
+    const saveVerb = String.raw`(?:spara|save|lagra|store|capture|anteckna|notera|skriv ner|arkivera|lägg\s+(?:till|in)|add|put|create.*note)`;
+    const retrieveVerb = String.raw`(?:berätta|tell|find|hitta|sök|search|show|visa|sammanfatta|summari[sz]e|relaterade|related|ge mig|give me|översikt|overview|vad jag vet|what i know)`;
     const isMixed =
-      /\b(spara|save|lagra|anteckna|notera|skriv ner|arkivera)\b.*\b(och|and)\b.*\b(berätta|tell|find|hitta|sök|search|sammanfatta|summari[sz]e|relaterade|related)\b/i.test(
-        text,
-      ) ||
-      /\b(berätta|tell|find|hitta|sök|search)\b.*\b(och|and)\b.*\b(spara|save|lagra)\b/i.test(text);
+      new RegExp(`\\b${saveVerb}\\b[\\s\\S]*\\b(?:och|and|also)\\b[\\s\\S]*\\b${retrieveVerb}\\b`, "i").test(text) ||
+      new RegExp(`\\b${retrieveVerb}\\b[\\s\\S]*\\b(?:och|and|also)\\b[\\s\\S]*\\b${saveVerb}\\b`, "i").test(text);
 
     if (isMixed) {
       label = "mixed";
       confidence = 0.82;
       rationale = "Save-and-retrieve in one message.";
     } else if (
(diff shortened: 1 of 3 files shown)


heimdallSWE (commented):
Solid prompt + scripted-mock improvements. The prompt rules-of-thumb and three new examples are coherent with each other, and the scripted-mock regexes in `eval/runner/classify.ts` have been updated in lockstep so the eval ground-truth aligns with the new prompt expectations.

## Bugs / concerns

**1. `[\s\S]*` in the mixed regex over-matches across full message length.**

Before:
```ts
/\b(save|...)\b.*\b(och|and)\b.*\b(berätta|...)\b/i
```
After:
```ts
new RegExp(`\\b${saveVerb}\\b[\\s\\S]*\\b(?:och|and|also)\\b[\\s\\S]*\\b${retrieveVerb}\\b`, "i")
```

`[\s\S]*` is greedy and unbounded across newlines. For a long message that contains, say, "save my draft" early and "tell me about" 200 chars later as a totally unrelated thought, the new regex still classifies mixed. The old `.*` (single-line) was less prone to this because the connector had to be on the same line. For Gemmera's chat UX this is probably fine — users rarely send multi-line save+ask in one turn — but worth confirming against the eval harness #140 that mixed precision hasn't dropped.

**2. Capturing vs non-capturing group inconsistency.**

```ts
// new saveVerb (non-capturing):
String.raw`(?:spara|save|...|lägg\s+(?:till|in)|add|...)`

// existing capture-path regex (capturing):
/\b(spara|...|lägg\s+(till|in)|skapa.*anteckning|...)\b/i
```

`.test()` doesn't care about capture, so this is purely stylistic, but the diff would be easier to scan if both used `(?:...)`. Optional cleanup.

## Smaller

- **Prompt grew ~300 tokens** (rules-of-thumb + 3 examples). For a local gemma 4B classifier that's non-trivial context. Suggest running #140's eval harness pre/post to confirm no token-budget regression and ideally a precision/recall lift on the new examples.
- **"What is the difference between saving and asking?" → meta** is the right call, but the LLM might be tempted to read this as `ask` (user is "asking"). The few-shot rationale clarifies — good. Could add a slightly tougher variant as a stress example, e.g. "What's the diff between mixed and capture for my prompts?" → still meta.
- **`how I delete a note?`** vs **`hur raderar`** — added the Swedish `raderar` and English `delete a note`. Good. Missing Swedish for "what is the difference" — `vad är skillnaden mellan`. Worth a follow-up; not blocking.

## Eval coverage

Worth adding the three new prompt examples to `eval/golden/fixtures/` (assuming that's where the classifier golden set lives — flag from a previous PR). Otherwise the scripted mock validates them but the real-model golden set won't catch a regression on them.

Approving the direction; would like #1 sanity-checked against eval harness #140 before merge.

PR #157, opened 2026-05-14
feat(tools): gate schemas by route intent

### LINKED ISSUE #149: feat(tools): gate tool schemas by route intent
## Problem

The chat loop currently sends all read and write tool schemas on every LLM turn. This is simple, but local Gemma spends prompt budget on tools that often cannot be relevant for the selected route.

## Acceptance criteria

- Choose tool schemas from the current route/intent instead of always sending `ALL_TOOLS`.
- Ask/query turns should prefer read tools.
- Capture/write turns should prefer write tools.
- Mixed turns should still get the tools needed for both phases.
- Unknown or low-confidence routes should remain safe and user-visible.
- Add tests for tool selection behavior.
- Run `npm run eval:classifier:ci` and `npm run eval:retrieval:ci` to check for regressions.

SimonProjekt:
## What
- Adds a route-aware tool schema selector.
- Sends read tools for ask turns, write tools for capture turns, and both tool sets for mixed turns.
- Withholds tools for meta, unknown, disambiguation, and classifier fallback routes.
- Adds focused tests for tool selection behavior.

## Why
Gemma should not spend prompt budget on tool schemas that cannot be relevant to the selected route. Unknown or low-confidence routes stay conservative until the user-visible route decision is resolved.

Closes #149

## How to test
- npm test -- --run src/services/tool-dispatcher.test.ts src/view.test.ts
- npm test
- npm run build
- npm run eval:retrieval:ci
- npm run eval:classifier:ci (currently fails on dev because #150 is not merged yet; #156 fixes the six existing classifier golden failures)
- npx tsc --noEmit (currently blocked on dev by #146; #154 fixes the NotePreviewModal Component owner issue)

### DIFF
--- a/src/services/tool-dispatcher.test.ts
+++ b/src/services/tool-dispatcher.test.ts
@@ -1,6 +1,16 @@
 import { describe, expect, it, vi } from "vitest";
+import type { RouteDecision } from "../contracts";
 import { MockVaultService } from "../contracts/mocks/mock-vault";
-import { dispatchToolCall, appendUnderDatedSection, patchFrontmatter, type ToolDispatchDeps } from "./tool-dispatcher";
+import {
+  ALL_TOOLS,
+  READ_TOOLS,
+  WRITE_TOOLS,
+  appendUnderDatedSection,
+  dispatchToolCall,
+  patchFrontmatter,
+  selectToolsForIntent,
+  type ToolDispatchDeps,
+} from "./tool-dispatcher";
 import { InMemoryJobQueue } from "./in-memory-job-queue";
 import { IngestWriter } from "./ingest-writer";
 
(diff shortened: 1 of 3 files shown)


heimdallSWE (commented):
Logic of `selectToolsForIntent` is correct and the tests are thorough, but the PR description doesn't match what the code actually does in this repo today, and that's worth confirming before merging.

**Dispatch graph.** `handleSend` routes `capture` → `runCapture` (uses `runIngest`), `mixed` → `runMixed` (uses `runIngest` + `runQuery`), `ask` → `runAsk` (uses `runQuery`). Only `meta` (and any unknown intent) falls through to `runChat`. `runChat` is also the only call site of `services.llm.chat({tools: ...})` in the codebase — every other LLM call (`runIngest`, `runQuery`, `classifier-llm`, `ingest-parser`, `ingest-strategy`) calls `.chat()` without a `tools` field.

So the wiring change at `view.ts:591` only affects the `meta` path, and `selectToolsForIntent("meta", route)` returns `[]`. Net effect:

- `meta` turns: tools change from `ALL_TOOLS` → `[]` ✅ (this is fine and matches the description's spirit — meta shouldn't need tools)
- `ask` / `capture` / `mixed`: no change, because those intents are routed away from `runChat` before this code path ever runs. The `case "ask"` / `case "capture"` / `case "mixed"` branches are dead from the only caller.

A couple of options:

1. If the intent is just to keep the selector ready in case a future caller (e.g. a unified turn runner) needs it, that's defense-in-depth — fine, but a one-line comment above `selectToolsForIntent` saying "currently called only from `runChat` (meta path); other intents route through their own orchestrators" would save the next reader 15 minutes.
2. If the PR description's "Sends read tools for ask turns, write tools for capture turns" is meant to describe live behavior, then there's a wiring step missing — `runAsk` / `runCapture` / `runMixed` don't pass tools to `.chat()` at all, so the selector for those intents isn't actually reached.

Could you confirm which one this is? Happy to file a follow-up issue if option 1; if option 2, the missing wiring is probably the bigger half of #149.

Two smaller things while I'm here:

- The PR is currently `CONFLICTING` against dev — `tool-dispatcher.ts` / `tool-dispatcher.test.ts` / `view.ts` all changed in #153 and #158 since this branch was cut. Will need a rebase.
- `selectToolsForIntent("meta", …)` and `selectToolsForIntent(null, …)` both return `[]` via the `default` arm — the explicit `case "meta":` falling into `default` is correct but easy to misread as "meta returns []  but unknown might return something else". Tiny stylistic nit; not blocking.

PR #161, opened 2026-05-14
feat(ui): NotePreviewModal split mode — N independent confirm flows (#160)

### LINKED ISSUE #160: feat(ui): NotePreviewModal — split mode (N independent confirm flows)
Split-out from #52, where the modal portion (single-spec commit gate) shipped via #153.

## Scope

When the ingest pipeline emits a `split` outcome with N `NoteSpec` candidates (see `IngestOutcome` in `src/contracts/ingest.ts`), the UI must surface N independent NotePreviewModal flows so the user can confirm, edit, or cancel each candidate separately.

### Acceptance (from original #52)

- Split mode produces N independent confirm flows.
- Cancelling one does not cancel the others.
- Confirming one writes that note; the remaining flows continue uninterrupted.
- Order of presentation is stable (deterministic by candidate index).

## Why deferred

No upstream caller passes >1 spec today — `propose_notes` doesn't exist yet and the ingest orchestrator currently routes all single-spec outcomes through the modal landed in #153. Implementing split mode without a caller would be dead code; it should land alongside the split-producing path.

## Implementation notes

- The dispatcher already wraps `openNotePreview` in a single Promise. Split mode needs a parallel-flow primitive (`Promise.allSettled` over N modals, or a queue-style "one at a time" UX — pick before implementing).
- The modal currently shows one body preview; for split mode the user needs to see *which* candidate they're confirming (title in the heading, "1 of N" indicator).
- Each cancellation should result in `{ kind: "cancelled", index }` so the orchestrator can keep the confirmed siblings.

## Tracking

Closes the remaining acceptance criterion of #52. Once shipped, this issue + #153 collectively close the original.

heimdallSWE:
Closes #160. Completes the remaining acceptance criterion of #52 (split-out from #153, which shipped the single-spec commit gate).

## Background

The ingest orchestrator has produced `{ kind: "split", candidates }` previews since `splitSpecByHeadings` landed in `ingest-strategy.ts` — but the UI adapter (`ingest-preview-modal`) only handled `dedup` and `append`. Split candidates had no surface. This PR adds the missing branch without touching the just-merged `note-preview-modal`.

## Approach

**Queue UX (one modal at a time)** over parallel modals: simpler, matches Obsidian conventions, splits the user's attention by candidate.

- New `src/ui/split-preview.ts` exports `openSplitPreview(app, candidates, defaults)`. Walks candidates in order, opening a `SplitPreviewModal` per candidate with a `"(i of N)"` heading and three buttons: **Save / Skip / Cancel all**.
- Reuses `validateNotePreview`, `NOTE_TYPES`, `NOTE_STATUSES` from `note-preview-modal` so the rag.md schema gate stays in one place.
- `reduceSplitOutcomes` is exported as a pure helper so orchestration logic is testable without DOM.
- `view.ts` routes via a new `routeIngestPreview(preview)` method — split with ≥2 candidates → `openSplitPreview`, everything else → existing `openIngestPreview`. Both call sites (capture and mixed) go through the same router.

## Acceptance (issue #160)

| Criterion | Status |
|---|---|
| Split mode produces N independent confirm flows | ✅ Sequential `SplitPreviewModal`, one per candidate |
| Cancelling one does not cancel the others | ✅ **Skip** advances without writing; **Cancel all** stops iteration and ships whatever was already confirmed |
| Confirming one writes that note; remaining flows continue | ✅ Confirmed specs accumulate, orchestrator writes them after the loop ends |
| Order of presentation is stable (deterministic by index) | ✅ `for (let i = 0; i < candidates.length; i++)`; verified in tests |

## Edge case: empty result

If the user skips everything or cancels before saving anything, `reduceSplitOutcomes` collapses to `{ action: "cancel" }` rather than `{ action: "split_confirm", confirmed: [] }`. Without this, the orchestrator's `split_saved` branch would fire a misleading "saved 0 notes" notice.

Esc / outside-click during split mode resolves to **skip-this-one**, never cancel-all, so a misclick can't drop confirmed siblings.

## Test plan

- [x] `npm test` → 746/746 (was 734; +12: split-preview helpers + reducer cases)
- [x] `npm run build` → clean
- [x] `tsc --noEmit` → only the pre-existing `MarkdownRenderer.render(..., this)` typing nit on dev (same nit `note-preview-modal.ts` carries); no new errors
- [ ] Manual: trigger a split outcome (paste content with multiple `## ` headings, dedup-decider returns `split`) → confirm modal shows `(1 of N)`, advances on Save, skips on Skip, ends loop on Cancel all
- [ ] Manual: cancel-all after saving 1 of 3 → vault has 1 new note, not 0 and not 3

## Out of scope

- `propose_notes` LLM tool (planning-doc only, not implemented).
- Re-ordering or merging candidates (not in AC).
- Rollback if write fails mid-sequence (orchestrator-side concern; current behaviour writes sequentially and fails on first error — separate issue if needed).

### DIFF
--- a/src/ui/split-preview.test.ts
+++ b/src/ui/split-preview.test.ts
@@ -1,0 +1,112 @@
+import { describe, expect, it } from "vitest";
+import type { NoteSpec } from "../contracts/ingest";
+import type { NotePreviewResult } from "./note-preview-modal";
+import { reduceSplitOutcomes, type OneOutcome } from "./split-preview";
+
+function spec(overrides: Partial<NoteSpec> = {}): NoteSpec {
+  return {
+    title: "Topic A",
+    type: "source",
+    tags: ["t1"],
+    aliases: [],
+    source: "chat-paste",
+    entities: [],
+    related: [],
+    status: "inbox",
+    summary: "Notes about Topic A.",
+    key_points: [],
+    body_markdown: "## Topic A\n\nbody",
+    cowork: {
+      source: "ingest",
+      run_id: "run-1",
+      model: "test-model",
+      version: "0.0.1",
+      confidence: "high",
+    },
+    ...overrides,
+  };
+}
+
+function result(overrides: Partial<NotePreviewResult> = {}): NotePreviewResult {
+  return {
+    confirmed: true,
+    title: "Edited title",
+    folder: "Inbox/",
+    type: "source",
+    status: "inbox",
+    tags: ["edited"],
+    aliases: [],
+    summary: "Edited summary.",
+    ...overrides,
+  };
+}
+
+const saved = (r?: Partial<NotePreviewResult>): OneOutcome => ({ kind: "saved", result: result(r) });
+const skipped: OneOutcome = { kind: "skipped" };
+const cancelled: OneOutcome = { kind: "cancelled" };
+
+describe("reduceSplitOutcomes — split-mode acceptance (#160)", () => {
+  it("returns split_confirm with every candidate when each is saved", () => {
+    const candidates = [spec({ title: "A" }), spec({ title: "B" }), spec({ title: "C" })];
+    const out = reduceSplitOutcomes(candidates, [
+      saved({ title: "A!" }),
+      saved({ title: "B!" }),
+      saved({ title: "C!" }),
+    ]);
+    expect(out.action).toBe("split_confirm");
+    if (out.action !== "split_confirm") return;
+    expect(out.confirmed.map((c) => c.title)).toEqual(["A!", "B!", "C!"]);
+  });
+
+  it("preserves order by candidate index", () => {
+    const candidates = [spec({ title: "first" }), spec({ title: "second" }), spec({ title: "third" })];
+    const out = reduceSplitOutcomes(candidates, [
+      saved({ title: "1" }),
+      skipped,
+      saved({ title: "3" }),
+    ]);
+    if (out.action !== "split_confirm") throw new Error("expected split_confirm");
+    expect(out.confirmed.map((c) => c.title)).toEqual(["1", "3"]);
+  });
+
+  it("skip preserves siblings — skipping one does not cancel the rest", () => {
+    const candidates = [spec({ title: "A" }), spec({ title: "B" })];
+    const out = reduceSplitOutcomes(candidates, [skipped, saved({ title: "B!" })]);
+    if (out.action !== "split_confirm") throw new Error("expected split_confirm");
+    expect(out.confirmed).toHaveLength(1);
+    expect(out.confirmed[0].title).toBe("B!");
+  });
+
+  it("cancel-all ships whatever was already confirmed", () => {
+    const candidates = [spec({ title: "A" }), spec({ title: "B" }), spec({ title: "C" })];
+    const out = reduceSplitOutcomes(candidates, [saved({ title: "kept" }), cancelled]);
+    if (out.action !== "split_confirm") throw new Error("expected split_confirm");
+    expect(out.confirmed).toHaveLength(1);
+    expect(out.confirmed[0].title).toBe("kept");
+  });
+
+  it("collapses to plain cancel when nothing was confirmed before cancel-all", () => {
+    const candidates = [spec({ title: "A" }), spec({ title: "B" })];
+    const out = reduceSplitOutcomes(candidates, [skipped, cancelled]);
+    expect(out.action).toBe("cancel");
+  });
+
+  it("collapses to plain cancel when every candidate is skipped (no writes)", () => {
+    const candidates = [spec({ title: "A" }), spec({ title: "B" })];
+    const out = reduceSplitOutcomes(candidates, [skipped, skipped]);
+    expect(out.action).toBe("cancel");
+  });
+
+  it("carries non-edited NoteSpec fields through unchanged (body, source, cowork)", () => {
+    const candidates = [
+      spec({ title: "orig", body_markdown: "## orig body", source: "chat-paste" }),
+    ];
+    const out = reduceSplitOutcomes(candidates, [saved({ title: "renamed" })]);
+    if (out.action !== "split_confirm") throw new Error("expected split_confirm");
+    const c = out.confirmed[0];
+    expect(c.title).toBe("renamed");
+    expect(c.body_markdown).toBe("## orig body");
+    expect(c.source).toBe("chat-paste");
+    expect(c.cowork.run_id).toBe("run-1");
+  });
+});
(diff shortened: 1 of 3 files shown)


SimonProjekt (changes requested):
## Review

Overall this is solid work — the `resolved` guard, Esc→skip UX, exported `reduceSplitOutcomes` for pure testing, and reuse of `validateNotePreview`/`NOTE_TYPES`/`NOTE_STATUSES` are all good calls. 12 new tests covering all branches of the reducer is appreciated.

Two things to address before merge:

---

### 1. Routing threshold `> 1` is likely wrong (`view.ts`)

```ts
if (preview.kind === "split" && preview.candidates && preview.candidates.length > 1) {
```

A split with exactly 1 candidate falls through to `openIngestPreview`. If that function doesn't handle `kind === "split"` the result is a silent no-op. Should probably be `>= 1`, or add a comment explaining why a single-candidate split is intentionally treated as non-split (and verify `openIngestPreview` handles it gracefully).

---

### 2. `applyEditsToSpec` silently drops `folder` (`split-preview.ts`)

```ts
function applyEditsToSpec(base: NoteSpec, edits: NotePreviewResult): NoteSpec {
  return { ...base, title: edits.title, type: edits.type, ... }; // folder missing
}
```

`NotePreviewResult` has a `folder` field the user can edit in the modal, but it's never copied back to `NoteSpec`. If the orchestrator derives the write path from `confirmed[i]` this edit is silently discarded. Please confirm how `ingest-orchestrator` consumes `{ action: "split_confirm", confirmed }` and whether `folder` needs to be on `NoteSpec` or passed separately.

---

### Minor / optional

**Code duplication** — `openSplitPreview` and `reduceSplitOutcomes` share the same loop + `applyEditsToSpec` logic. Consider refactoring `openSplitPreview` to collect outcomes into an array and then call `reduceSplitOutcomes`, so the tested code path covers production exactly:

```ts
export async function openSplitPreview(...): Promise<PreviewDecision> {
  const outcomes: OneOutcome[] = [];
  for (let i = 0; i < candidates.length; i++) {
    const o = await openOne(app, candidates[i], defaults.folder, i, candidates.length);
    outcomes.push(o);
    if (o.kind === "cancelled") break;
  }
  return reduceSplitOutcomes(candidates, outcomes);
}
```

**Manual test plan items are unchecked** — confirm the two Obsidian manual tests have been verified before merge.

heimdallSWE (commented):
Great architecture — `reduceSplitOutcomes` as a pure helper is exactly the right separation; the seven unit tests cover all branch combinations, and routing through `routeIngestPreview()` keeps capture and mixed flows DRY. Two concrete bugs to address plus a couple of UX choices worth confirming.

## Bugs

**1. Will fail `tsc --noEmit` once rebased onto current `dev`.**

```ts
MarkdownRenderer.render(this.app, this.candidate.body_markdown, previewEl, "", this).catch(...)
```

The fifth arg requires a `Component`; `this` is a `Modal`. #154 (just merged) fixed this exact pattern in `note-preview-modal.ts` by adding a `renderOwner = new Component()` field with explicit `load()` in `onOpen` and `unload()` in `onClose`. This PR was branched before #154 landed, so its CI workflow doesn't include the typecheck step — but once rebased it'll fail. Mirror the #154 fix here.

**2. User-edited folder is silently dropped.**

`SplitPreviewModal` shows a `Folder` input and the validator accepts it (with fallback to `defaults.folder`), but `applyEditsToSpec` only carries `title / type / status / tags / aliases / summary` forward — `folder` is left behind. The orchestrator then writes via:

```ts
await deps.writer.writeNew(candidate, { folder: deps.inboxFolder });  // always default
```

Net: if the user types a different folder in any per-candidate modal, their input is discarded. Two ways to fix:

- **Drop the Folder input** from split-preview modals and rely on `deps.inboxFolder`, with a brief note in the heading ("Saving to `Inbox/`"). Simpler — split notes are auto-derived from a single ingest message, so per-candidate folder is unusual.
- **Plumb folder through.** Extend `applyEditsToSpec` and the orchestrator's split branch to honor `confirmed[i].folder` if you want per-candidate placement.

Either's fine; pick one and the test ("carries non-edited NoteSpec fields through unchanged") should grow an assertion on whichever you chose.

## UX

**3. Esc / outside-click means "skip this one" — divergent from `NotePreviewModal`.**

`note-preview-modal.ts` treats Esc as Cancel (null → "Save cancelled"). `split-preview.ts` treats Esc as Skip and silently advances:

```ts
onClose(): void {
  if (!this.resolved) {
    // Esc / outside-click during split mode means "skip this one"
    this.onDone({ kind: "skipped" });
  }
}
```

The inline comment frames this as deliberate, but the user has no visual cue. After two "Esc-to-cancel" instincts they'll have skipped two candidates and won't realize until they see the write summary. Options:

- Add a one-line hint under the "i of N" heading: *"Esc skips this candidate. Use Cancel all to abort."*
- Or align with single-mode: Esc = cancel all. Then the explicit Skip button is the only way to skip.

I'd lean **align with single-mode** — Esc semantics should be predictable across modal types in the same plugin.

## Smaller

**4. CSS class `gemmera-split-preview` is added but unstyled.** No `styles.css` change in this PR. The class is currently a no-op hook. Either drop it for now or add a minimal rule (e.g., a subtle progress indicator color so the user can visually distinguish split-mode from single-mode at a glance).

**5. Empty summary on split candidates surprises the Save gate.**

If `splitSpecByHeadings` produces a candidate with an empty `summary`, the Save button starts disabled and the user can't proceed until they type something. That's correct per the schema, but it's not obvious from the UI why Save is disabled before they've touched any field. A `placeholder` on the summary textarea ("Required: 1–2 sentences summarizing this section") would close the loop.

**6. `OneOutcome` is exported just for tests.** Fine but slightly leaky — alternatively, export a factory like `splitOutcomeSaved(result)` / `splitOutcomeSkipped()` / `splitOutcomeCancelled()` so tests don't need to know the discriminant shape. Not blocking.

## Praise

- **`reduceSplitOutcomes` as a pure helper** is the right testing seam — coverage now spans every save/skip/cancel combination without DOM. Bar set high.
- **Single `routeIngestPreview` in view.ts** keeps the split branch invisible to both capture and mixed callers. Clean refactor.
- **Reusing `validateNotePreview`, `NOTE_TYPES`, `NOTE_STATUSES`** keeps the schema gate canonical — exactly the right reuse pattern.
- **Queue UX over parallel modals.** Matches Obsidian conventions and respects user attention; the only reasonable choice.
- **Test coverage on "carries non-edited NoteSpec fields through unchanged"** is the kind of regression test I'd want to see in every refactor.

**Priority:** Bug #1 (typecheck) is a hard CI blocker once rebased. Bug #2 (folder dropped) is silent so it'd ship undetected; worth fixing or explicitly removing the input. Bug #3 (Esc semantics) is a judgment call.

PR #162, opened 2026-05-14
feat(ui): inline preview in wide context panel (#55)

### LINKED ISSUE #55: Direct-save path + Notice with Undo (preview-off mode)
When "Always preview before save" is off, Cowork writes directly to the configured inbox folder with the proposed content and surfaces a Notice with an Undo action. Undo reverts the write via `vault.trash`.

Also implement the wide-mode "inline preview" setting (default off) that collapses the modal step and surfaces the same preview form in the right context panel.

### Acceptance
- Toggling preview-off and saving writes the file and shows the Notice with Undo.
- Undo trashes the file via `vault.trash` (recoverable from .trash).
- Inline-preview mode in wide layout shows the form in the context panel and never opens a modal; in narrow layout it falls back to the modal.

### Reference
`planning/ui-surfaces.md` — "Preview modal (pre-save confirmation)" final two paragraphs.

heimdallSWE:
Closes #55. Final UI-surfaces v1 piece.

## Background

The issue's three AC items map to existing + missing work on `dev`:

| AC | Status |
|---|---|
| Toggling preview-off and saving writes the file + shows the Notice with Undo | ✅ Already on dev — `ingest-orchestrator.ts:173-176` skips preview for high-confidence creates when `alwaysPreview === false`; `notices.ts:showSaveUndoNotice` fires for create-mode saves. |
| Undo trashes via `vault.trash` (recoverable from `.trash`) | ✅ Already on dev — `notices.ts:12` calls `app.vault.trash(file, false)`. |
| **Inline-preview mode in wide layout: form in panel, no modal; narrow falls back to modal** | ❌ This PR |

## Approach

- New setting `inlinePreviewInWidePanel` (default off). Settings-tab toggle right under "Always preview before save."
- `ContextPanel` grows an `inline-preview` state. `setInlinePreview(preview, fallbackFolder, onDecide)` renders the same editable form as `NotePreviewModal` (title / folder / type / status / tags / aliases / summary), reusing `validateNotePreview` so the rag.md schema gate stays in one place.
- `ContextPanel.isWideLayoutActive()` reads visibility from the DOM (`offsetParent !== null`) so JS stays in sync with the CSS `@container gemmera (min-width: 600px)` rule that hides the panel in narrow mode — no duplicate breakpoint.
- `view.ts.routeIngestPreview(preview)` decides per-call: inline when the setting is on AND panel is visible AND `preview.kind === "save"`; modal otherwise. Append, dedup-ask, and split keep their modals — they aren't editable forms.
- The inline **Save** returns `{action: "edit", spec}` through the orchestrator's existing edit channel; a one-shot `inlinePreviewAutoConfirm` flag short-circuits the next preview call to `{action: "confirm"}` so the user clicks Save once, not twice.
- `mergeEditsIntoSpec` is exported and unit-tested — the field-merge contract is verified without spinning up DOM.

## Acceptance (issue #55)

| Criterion | Status |
|---|---|
| Toggling preview-off + saving writes the file + Notice with Undo | ✅ Pre-existing on dev (see Background) |
| Undo trashes via `vault.trash` | ✅ Pre-existing on dev |
| Inline-preview in wide layout shows form in panel, never opens modal | ✅ `routeIngestPreview` skips `openIngestPreview` when `isWideLayoutActive()` |
| In narrow layout, falls back to modal | ✅ `isWideLayoutActive()` returns false when `offsetParent === null` (CSS `display: none`); `routeIngestPreview` falls through to `openIngestPreview` |

## Test plan

- [x] `npm test` → 741/741 (was 734; +7: `mergeEditsIntoSpec` cases on top of existing context-panel tests)
- [x] `npm run build` → clean
- [x] `tsc --noEmit` → only the pre-existing `MarkdownRenderer.render(..., this)` typing nit; no new errors
- [ ] Manual: toggle "Inline preview in wide panel" on; widen pane >600 px → capture turn renders form in panel, no modal opens
- [ ] Manual: narrow pane <600 px with setting on → capture turn opens modal (panel is hidden)
- [ ] Manual: with setting off, capture always uses modal regardless of width
- [ ] Manual: inline Save then immediate ingest → Notice with Undo fires, file is at expected path

## Out of scope

- Inline preview for append / dedup-ask / split — those have non-form button dialogs and would clutter the panel column. Keeping them in the modal preserves visual focus.

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -21,6 +21,14 @@ export interface GemmeraSettings {
    */
   alwaysPreviewBeforeSave: boolean;
 
+  /**
+   * When true (and the chat pane is wide enough to show the context panel),
+   * the preview form renders inline in the right-side panel instead of
+   * opening a modal. Falls back to the modal in narrow layout. Default off.
+   * Wired to Settings → "Inline preview in wide panel" (#55).
+   */
+  inlinePreviewInWidePanel: boolean;
+
   /** Folder for new ingest notes. Default `Inbox/`. */
   inboxFolder: string;
 
@@ -76,6 +84,7 @@ export const DEFAULT_SETTINGS: GemmeraSettings = {
   llmBackend: "ollama",
   showClassifierDecisions: false,
   alwaysPreviewBeforeSave: true,
+  inlinePreviewInWidePanel: false,
   inboxFolder: "Inbox/",
   dedupThreshold: 0.85,
   ollamaPathMode: "auto",
--- a/src/ui/context-panel.test.ts
+++ b/src/ui/context-panel.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it } from "vitest";
-import { MAX_SNIPPET_CHARS, snippet, whyLabel } from "./context-panel";
+import type { NoteSpec } from "../contracts/ingest";
+import { MAX_SNIPPET_CHARS, mergeEditsIntoSpec, snippet, whyLabel } from "./context-panel";
 
 describe("whyLabel", () => {
   it("maps each winning signal to a short tag", () => {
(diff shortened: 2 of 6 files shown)


SimonProjekt (commented):
Good overall design — `routeIngestPreview` / `shouldUseInlinePreview` are clean, `mergeEditsIntoSpec` is well-tested, and `isWideLayoutActive()` via `offsetParent` is a smart way to stay in sync with the CSS container query without a duplicate JS breakpoint.

Three bugs to fix before merge:

## Bugs

**1. CSS class leaks on state transition** *(blocking)*

`renderInlinePreview` adds `.gemmera-context-content--inline-preview` to `root`, but the other three render functions (`renderIdle`, `renderQuery`, `renderIngestion`) don't remove it when they render. Transitioning from `inline-preview` → `idle` leaves the class on the root and the inline-preview CSS rules keep applying.

Fix: add `root.removeClass("gemmera-context-content--inline-preview")` inside the other render functions, or handle it centrally in `render()` before the switch.

**2. `inlinePreviewAutoConfirm` not cleared on abort** *(medium)*

If the `edit` decision is returned but the orchestrator throws before making the second `preview()` call, the flag stays `true`. The next save turn auto-confirms immediately without showing a form — silent data loss risk.

Fix: reset the flag in `runCapture`/`runMixed`'s `finally` block, or at minimum reset it on any non-`edit` decision in `openInlinePreview`:

```ts
this.contextPanel?.setInlinePreview(preview, this.settings.inboxFolder, (decision) => {
  this.inlinePreviewAutoConfirm = decision.action === "edit";
  resolve(decision);
});
```
…plus `this.inlinePreviewAutoConfirm = false` in `runCapture`/`runMixed` `finally`.

**3. Folder field shown but silently ignored** *(medium)*

`collectRaw()` collects `folderInput.value` and passes it to `validateNotePreview`, but `mergeEditsIntoSpec` has no `folder` parameter — the folder value never reaches `{ action: "edit", spec }`. The user edits the folder field and nothing happens.

Either remove the folder field from the inline form (it's controlled by the `inboxFolder` setting anyway), or propagate it through `PreviewDecision`.

## Smaller

- **Verbose type annotation in settings-tab.ts** — `.addToggle((toggle: { setValue: ... onChange: ... }) => {` is inconsistent with every other toggle in the file, which just use `(toggle)` with type inference.
- **No Enter-to-save** in the inline form — the modal has `keydown` listeners for Enter; the inline form doesn't, which gives inconsistent keyboard UX.
- **`isWideLayoutActive()` untested** — JSDOM doesn't support `offsetParent` properly, so it makes sense not to test it, but a short comment in the test file (or on the method) would save the next reader from wondering why coverage is missing.

